PR #47, opened 2014-10-29
elm logger

sonjakhan:
moved this from the Logging repo (aspnet/Logging#38), removed the Mvc dependency from the sample, and fixed scoping


borgdylan:
@SonjaKhan I am trying to use elm but I am getting exceptions that the Form could not be read. Why should there be a form if i am accessing /Elm using the browser address bar directly?

Eilon:
@borgdylan can you log a bug for this? Sonja was an intern and isn't currently working on ELM (though we welcome any future contributions from her! - Hi, @SonjaKhan !).

Issue #66, opened 2014-12-10
Allow simple custom error pages for specific HTTP status codes


DamianEdwards:
IIS provides a mechanism for configuring custom pages/files to be shown when specific HTTP status codes occur, e.g. show 404.html when a 404 happens.

We should have a similar feature that works cross-server and is provided by the diagnostics package.

Tratcher:
A few things we did decide on:
- The primary purpose is to generate human readable HTML pages, preferably something that matches the rest of the site's style.
- It needs to be configurable per status code. The default behavior for any given status code is open to discussion.
- Only handle responses with no body, do not support replacing the body.
- A simple scenario would do a client side redirect to another URL responsible for rendering the page (with a 200 status code).
- Note: Responses should be greater than 500 bytes (?), IE will not show small error responses.

Outstanding Issues:

How should you generate the body?
- There are lots of problems with trying to re-execute the request to let MVC or static files generate the body.
  - They'll overwrite the status code
  - They may try to honor request cache headers (if-not-match) with strange side-effects
  - They may generate response cache headers. Caching errors is bad.

How can you avoid breaking an API request?
- Do you need to check the Accept header to avoid sending HTML to client expecting XML/JSON?
- For the redirect-to-page scenario, how do we suppress that for API calls?

blowdart:
Surely an API request should ignore the body and just go off status code, otherwise they'd have barfed on IIS anyway :)

Praburaj:
Planning to finish this coming week.

Praburaj:
This change went in today.

PR #83, opened 2015-01-27
StatusCodePages middleware

Tratcher:
#66

Submitting for design feedback, it's not ready for a detailed code review yet. Are the handlers provided useful (response, redirect, tangent, re-execute)?

Entropy MVC sample: https://github.com/aspnet/Entropy/pull/38

Re-execute works OK with MVC. Static files would only work well with the redirect scenario due to all the request and response headers it uses, and the fact that it overwrites the status code. However you could do a simpler version of static files by going directly to the WebRootFileProvider and serving the content yourself.

@loudej @davidfowl @GrabYourPitchforks @yishaigalatzer


src/Microsoft.AspNet.Diagnostics/ReasonPhraseHelper.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    // TODO: Move this into Abstractions
```
davidfowl: Yes please :smile:

Tratcher: https://github.com/aspnet/HttpAbstractions/pull/180

Tratcher:
https://github.com/aspnet/HttpAbstractions/pull/180

Tratcher:
All right, if nobody has any design feedback, how about normal code review feedback?

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+            return options.WithHandler(context =>
+            {
+                var body = string.Format(CultureInfo.InvariantCulture, bodyFormat, context.HttpContext.Response.StatusCode);
+                return context.HttpContext.Response.SendAsync(body, contentType);
+            });
+        }
+
+        /// <summary>
+        /// Specifies that responses should be handled by redirecting with the given location url template.
+        /// This may include a '{0}' placeholder for the status code. Ulrs starting with '~' will have PathBase prepeneded,
```
Eilon: Tpyo `Ulrs` --> `URLs`

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// where any other url will be used as is.
```
Eilon: and `url` --> `URL`

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Specifies an alternate middleware pipeline to execute to generate the response body.
+        /// </summary>
+        /// <param name="options"></param>
+        /// <param name="configuration"></param>
+        /// <returns></returns>
+        public static StatusCodePagesOptions WithTangent(this StatusCodePagesOptions options, Action<IApplicationBuilder> configuration)
```
Eilon: I'm not sure if I was in the design meetings for this, but this name is... weak :frowning: I'd rather even call it something like `WithPipeline` or something because that's at least closer to what it is.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            var builder = new ApplicationBuilder(serviceProvider: null); // TODO: services
```
Eilon: `TODO` --> log a bug?

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly RequestDelegate _next;
+        private readonly StatusCodePagesOptions _options;
+
+        public StatusCodePagesMiddleware(RequestDelegate next, StatusCodePagesOptions options)
+        {
+            _next = next;
+            _options = options;
+            if (_options.HandleAsync == null)
+            {
+                throw new ArgumentException("Missing HandleAsync implementation.");
```
Eilon: Can also pass in the offending parameter name, `options`.

Eilon:
Minus a few small cleanup comments, this looks :shipit: to me!

samples/StatusCodePagesSample/Startup.cs
```diff
@@ -0,0 +1,75 @@
...
+                var requestedStatusCode = context.Request.Query["statuscode"];
+                if (!string.IsNullOrEmpty(requestedStatusCode))
+                {
+                    context.Response.StatusCode = int.Parse(requestedStatusCode);
+                    return Task.FromResult(0);
+                }
+
+                var builder = new StringBuilder();
+                builder.AppendLine("<html><body>");
+                builder.AppendLine("<a href=\"" + context.Request.PathBase + "/missingpage/\">" + context.Request.PathBase + "/missingpage/</a><br>");
```
yishaigalatzer: should the pathbase be HTML encoded?

yishaigalatzer: better yet - should this execute a precompiled razor page instead?

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
yishaigalatzer: :+1:

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+        public async Task Invoke(HttpContext context)
+        {
+            await _next(context);
+
+            // Do nothing if a response body has already been provided.
+            if (context.Response.HeadersSent
+                || context.Response.StatusCode < 400
+                || context.Response.StatusCode >= 600
+                || context.Response.ContentLength.HasValue
+                || !string.IsNullOrEmpty(context.Response.ContentType))
```
yishaigalatzer: Is the explicit mechanism to turn the status pages off, is to set ContentLength to Zero? Would like to see a sample for an API style controller that turns the status code off. (Perhaps a resource filter that looks at the result on the way out and turns it off).

yishaigalatzer: Where is the handling for a buffered stream that was written to by downstream component.

e.g. MVC Set an HTML page with 404 - in this case response headers didn't go out, and something is in the stream already

samples/StatusCodePagesSample/Startup.cs
Eilon: Yes, definitely HTML encoded, good catch. Regarding Razor, yes, I think that Chris just didn't want to cram that into this PR (I think he mentioned that somewhere but I can't find it).

Tratcher: We don't want to over complicate the sample with a pre-compiled razor view, but we do want to put one in as the default behavior here: https://github.com/aspnet/Diagnostics/pull/83/files#diff-cab524a96dcf6a9dbfe1ea56f57e8350R18

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
Tratcher: You can bypass the middleware by doing any of the following:
- Response.Body.Flush();
- Response.ContentLength = 0;
- Response.ContentType = "text/plain";
- Set any status code < 400.

This middleware has no knowledge of response buffering. The only response buffering I'm aware of is in Helios, and that's just a write-behind buffer so it still sets HeadersSent on the first Write.

Tratcher:
@Praburaj Can you pull the code from my branch and take over this feature?

FYI @muratg

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
yishaigalatzer: would be nice to have a single thing you can call to turn it off, so it can be slapped in `ResourceFilter` for api controllers.

Is the logic as simple as?

```C#
if (StatusCode >= 400)
{
   Respons.Body.Flush();
}
```

Because looks like any other way, actually starts sending stuff that is not necessary. Can we reconsider having an explicit turn-off feature?

Tratcher: That logic would work but would always result in a chunked response. Setting Content-Length: 0 would be better if you know it's going to be empty. The servers would have set the content-length to 0 anyways if the request unwound without sending anything so you're not adding anything extra to the client, just setting it sooner.

An explicit feature is possible, but I don't think the behavior is interesting enough yet to warrant one.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
Tratcher: You need access to the original app builder to get the services (and use `builder.New()`). I think the pattern for all of these should be re-worked like this:
```c#
public static IApplicationBuilder UseStatusCodePagesWithTangent(this IApplicationBuilder app, Action<IApplicationBuilder> configuration)
```

Eilon: Ah, interesting... still worth logging a bug, or otherwise I doubt we'd remember to do this ever...

muratg:
Looks good apart from the comments above. :shipit: 

@Praburaj please feel free to take this over.

Praburaj:
@Tratcher @muratg sure.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
Praburaj: Addressed this. Changed all of these extensions to `IApplicationBuilder` rather than that of `StatusCodePagesOptions`.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
Praburaj: Done.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
Praburaj: Fixed.

samples/StatusCodePagesSample/Startup.cs
Praburaj: HtmlEncoded the string.

Praburaj:
I have addressed all of the PR comments. I'm going to file a work item to put in a pre-compiled razor view for the default handler as soon as the feature goes in. @Eilon @muratg @Tratcher - let me know if you have any comments on these changes.

Praburaj:
Updated the PR as well.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        /// <summary>
+        /// Adds a StatusCodePages middleware with the specified handler that checks for responses with status codes 
+        /// between 400 and 599 that do not have a body.
+        /// </summary>
+        /// <param name="app"></param>
+        /// <param name="handler"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder WithHandler(this IApplicationBuilder app, Func<StatusCodeContext, Task> handler)
```
Praburaj: Updating the extension names to be same as `UseStatusCodePages`.

Eilon:
Looks :shipit: to me.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        /// <summary>
+        /// Adds a StatusCodePages middleware to the pipeline. Specifies that the response body should be generated by 
+        /// re-executing the request pipeline using an alternate path. This path may contain a '{0}' placeholder of the status code.
+        /// </summary>
+        /// <param name="app"></param>
+        /// <param name="pathFormat"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder WithReExecute(this IApplicationBuilder app, string pathFormat)
```
Tratcher: Missed one rename.

Praburaj:
Thanks @Eilon. Will merge it in.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
Praburaj: This has the same signature as another extension. So I left it with this name.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        /// <summary>
+        /// Adds a StatusCodePages middleware with the specified response body to send. This may include a '{0}' placeholder for the status code.
+        /// The middleware checks for responses with status codes between 400 and 599 that do not have a body.
+        /// </summary>
+        /// <param name="app"></param>
+        /// <param name="contentType"></param>
+        /// <param name="bodyFormat"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseStatusCodePages(this IApplicationBuilder app, string contentType, string bodyFormat)
```
Tratcher: Many of these overloads have similar parameters so it can be hard to pick the right one just by signature.  How about using a pattern like 'UseStatusCodePagesWithRedirects' to disambiguate?

Praburaj: That's a good point. That would solve the last name issue as well. Let me update.

Praburaj: Done.

Praburaj: @Tratcher any other comments?

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
yishaigalatzer: This was not addressed. Please log a follow up issue, this is not easy to write code that interacts with this middleware from MVC

PR #88, opened 2015-02-11
Adding ability to turn off the status code pages

Praburaj:
Added a feature named IStatusCodeFeature with an enabled property. MVC or any framework can get this feature and turn it off if
it does not want status code pages logic to take over and send HTML response.

Also refactored the sample flow to make it easier to understand.

@yishaigalatzer @Tratcher @Eilon @muratg

### DIFF
--- a/samples/StatusCodePagesSample/Startup.cs
+++ b/samples/StatusCodePagesSample/Startup.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
 using System.Net;
 using System.Text;
 using System.Threading.Tasks;
@@ -25,53 +26,74 @@ namespace StatusCodePagesSample
             // app.UseStatusCodePagesWithReExecute("/errors/{0}");
 
             // "/[?statuscode=400]"
-            app.Use((context, next) =>
+            app.Use(async (context, next) =>
             {
-                if (context.Request.Path.HasValue && !context.Request.Path.Equals(new PathString("/")))
-                {
-                    return next();
-                }
-
                 // Check for ?statuscode=400
                 var requestedStatusCode = context.Request.Query["statuscode"];
                 if (!string.IsNullOrEmpty(requestedStatusCode))
                 {
                     context.Response.StatusCode = int.Parse(requestedStatusCode);
-                    return Task.FromResult(0);
+
+                    // To turn off the StatusCode feature - For example the below code turns off the StatusCode middleware 
+                    // if the query contains a disableStatusCodePages=true parameter.
+                    var disableStatusCodePages = context.Request.Query["disableStatusCodePages"];
+                    if (disableStatusCodePages == "true")
+                    {
+                        var statusCodePagesFeature = context.GetFeature<IStatusCodePagesFeature>();
+                        if (statusCodePagesFeature != null)
+                        {
+                            statusCodePagesFeature.Enabled = false;
+                        }
+                    }
+
+                    await Task.FromResult(0);
                 }
-
-                var builder = new StringBuilder();
-                builder.AppendLine("<html><body>");
-                builder.AppendLine("<a href=\"" +
-                    WebUtility.HtmlEncode(context.Request.PathBase.ToString()) + "/missingpage/\">" +
-                    WebUtility.HtmlEncode(context.Request.PathBase.ToString()) + "/missingpage/</a><br>");
-
-                for (int statusCode = 400; statusCode < 600; statusCode++)
+                else
                 {
-                    builder.AppendLine("<a href=\"?statuscode=" + statusCode + "\">" + statusCode + "</a><br>");
+                    await next();
                 }
-                builder.AppendLine("</body></html>");
-                return context.Response.SendAsync(builder.ToString(), "text/html");
             });
 
             // "/errors/400"
-            app.Use((context, next) =>
+            app.Map("/errors", error =>
             {
-                PathString remainder;
-                if (context.Request.Path.StartsWithSegments(new PathString("/errors"), out remainder))
+                error.Run(async context =>
                 {
                     var builder = new StringBuilder();
                     builder.AppendLine("<html><body>");
-                    builder.AppendLine("An error occurred, Status Code: " + WebUtility.HtmlEncode(remainder.ToString().Substring(1)) + "<br>");
+                    builder.AppendLine("An error occurred, Status Code: " + WebUtility.HtmlEncode(context.Request.Path.ToString().Substring(1)) + "<br>");
                     var referrer = context.Request.Headers["referer"];
                     if (!string.IsNullOrEmpty(referrer))
                     {
                         builder.AppendLine("<a href=\"" + WebUtility.HtmlEncode(referrer) + "\">Retry " + WebUtility.HtmlEncode(referrer) + "</a><br>");
                     }
                     builder.AppendLine("</body></html>");
-                    return context.Response.SendAsync(builder.ToString(), "text/html");
+                    await context.Response.SendAsync(builder.ToString(), "text/html");
+                });
+            });
+
+            app.Run(async context =>
+            {
+                // Generates the HTML with all status codes.
+                var builder = new StringBuilder();
+                builder.AppendLine("<html><body>");
+                builder.AppendLine("<a href=\"" +
+                    WebUtility.HtmlEncode(context.Request.PathBase.ToString()) + "/missingpage/\">" +
+                    WebUtility.HtmlEncode(context.Request.PathBase.ToString()) + "/missingpage/</a><br>");
+
+                var space = string.Concat(Enumerable.Repeat("&nbsp;", 12));
+                builder.AppendFormat("<br><b>{0}{1}{2}</b><br>", "Status Code", space, "Status Code Pages");
+                for (int statusCode = 400; statusCode < 600; statusCode++)
+                {
+                    builder.AppendFormat("{0}{1}{2}{3}<br>",
+                        statusCode,
+                        space + space,
+                        string.Format("<a href=\"?statuscode={0}\">[Enabled]</a>{1}", statusCode, space),
+                        string.Format("<a href=\"?statuscode={0}&disableStatusCodePages=true\">[Disabled]</a>{1}", statusCode, space));
                 }
-                return next();
+
+                builder.AppendLine("</body></html>");
+                await context.Response.SendAsync(builder.ToString(), "text/html");
             });
         }
     }
--- a/src/Microsoft.AspNet.Diagnostics.Elm/project.json
+++ b/src/Microsoft.AspNet.Diagnostics.Elm/project.json
@@ -2,8 +2,7 @@
     "version": "1.0.0-*",
     "dependencies": {
         "Microsoft.AspNet.Diagnostics": "1.0.0-*",
-        "Microsoft.Framework.Logging.Interfaces": { "version": "1.0.0-*", "type": "build" },
-        "Microsoft.Framework.Runtime.Interfaces": { "version": "1.0.0-*", "type": "build" }
+        "Microsoft.Framework.Logging.Interfaces": "1.0.0-*"
     },
 
     "frameworks": {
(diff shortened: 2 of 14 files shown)


src/Microsoft.AspNet.Diagnostics/StatusCodeFeature.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    public class StatusCodeFeature : IStatusCodeFeature
+    {
+        public StatusCodeFeature()
+        {
+            Enabled = true;
```
Eilon: I think you can initialize this on the property itself in C# 6, no?

Eilon: public bool Enabled { get; set; } = true;

Praburaj: yup. Will fix it.

src/Microsoft.AspNet.Diagnostics/IStatusCodeFeature.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.Runtime;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    [AssemblyNeutral]
+    public interface IStatusCodeFeature
```
Tratcher: IStatusCodePagesFeature?

src/Microsoft.AspNet.Diagnostics/IStatusCodeFeature.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.Runtime;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    [AssemblyNeutral]
```
yishaigalatzer: AssemblyNeutral is going away, don't use it

src/Microsoft.AspNet.Diagnostics/StatusCodeFeature.cs
yishaigalatzer: Or go with Disabled

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
```diff
@@ -25,8 +25,17 @@ public StatusCodePagesMiddleware(RequestDelegate next, [NotNull]StatusCodePagesO
 
         public async Task Invoke(HttpContext context)
         {
+            context.SetFeature<IStatusCodeFeature>(new StatusCodeFeature());
+
             await _next(context);
 
+            var statusCodeFeature = context.GetFeature<IStatusCodeFeature>();
+            if (statusCodeFeature != null && !statusCodeFeature.Enabled)
```
yishaigalatzer: Since this middleware is setting this feature, shouldn't it be considered off if it is not around?

samples/StatusCodePagesSample/Startup.cs
```diff
@@ -25,53 +25,67 @@ public void Configure(IApplicationBuilder app)
...
-                }
 
-                var builder = new StringBuilder();
-                builder.AppendLine("<html><body>");
-                builder.AppendLine("<a href=\"" +
-                    WebUtility.HtmlEncode(context.Request.PathBase.ToString()) + "/missingpage/\">" +
-                    WebUtility.HtmlEncode(context.Request.PathBase.ToString()) + "/missingpage/</a><br>");
+                    // To turn off the StatusCode feature - For example the below code turns off the StatusCode middleware 
+                    // for even status codes for demonstration.
+                    if (context.Response.StatusCode % 2 == 0)
```
Tratcher: This is really random, just pick a specific status code to disable.

src/Microsoft.AspNet.Diagnostics/StatusCodeFeature.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    public class StatusCodeFeature : IStatusCodeFeature
+    {
+        public StatusCodeFeature()
```
yishaigalatzer: How is MVC consuming it? What does it have to take dependency on, now that there is no `[AssemblyNeutral]` support

yishaigalatzer:
:watch:

src/Microsoft.AspNet.Diagnostics/IStatusCodeFeature.cs
Tratcher: Really? What's the replacement?

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
Praburaj: good point.

src/Microsoft.AspNet.Diagnostics/IStatusCodeFeature.cs
Praburaj: yup.

src/Microsoft.AspNet.Diagnostics/StatusCodeFeature.cs
Praburaj: Like how rest other stack does it. I don't have an answer on this yet :)

samples/StatusCodePagesSample/Startup.cs
Praburaj: Because the HTML page generated with status codes is long with `200` entries it would be hard if I just disable it on one status code. Alternative status code did not look random to me as there is a pattern. Alternatively I can generate links in the HTML page for both status code pages enabled as well as disabled. How does that look?

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
Tratcher: So you could disable the middleware by just removing the feature interface? That makes the Enabled property redundant. Hmm, that would reduce that section of the sample to `context.SetFeature<IStatusCodeFeature>(null);`

Is there any benefit to knowing that the application is capable of this feature, but that it is currently disabled?

Praburaj: >Is there any benefit to knowing that the application is capable of this feature, but that it is currently disabled?

Currently nothing unless we add more to the feature interface itself which will require such a check.

Tratcher: Most of this ambiguity is coming from the fact that you put something in the feature collection and then check if it's still there just a few lines later. This assigns redundant meaning to it's removal.

I think this impl would make more sense if you stored the feature object in a local variable and just rely on the value of `Enabled`.

Praburaj:
@yishaigalatzer @Tratcher Updated the PR.

Tratcher:
Woops, I think you just missed my last comment.

Praburaj:
@Tratcher - which one?

Tratcher:
https://github.com/aspnet/Diagnostics/pull/88#discussion_r24526431

Praburaj:
i removed the enabled property totally and just checking for the presence of the capability. If there are no concerns here, I can push this.

src/Microsoft.AspNet.Diagnostics/IStatusCodeFeature.cs
yishaigalatzer: No replacement

yishaigalatzer:
Not ready, ANI was not removed and still not clear how we can add an MVC filter without taking dependency on StatusPages directly.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesFeature.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    public class StatusCodePagesFeature : IStatusCodePagesFeature
```
Eilon: Add a doc comment for this.

src/Microsoft.AspNet.Diagnostics/StatusCodePagesMiddleware.cs
```diff
@@ -25,8 +25,16 @@ public StatusCodePagesMiddleware(RequestDelegate next, [NotNull]StatusCodePagesO
 
         public async Task Invoke(HttpContext context)
         {
+            context.SetFeature<IStatusCodePagesFeature>(new StatusCodePagesFeature());
+
             await _next(context);
 
+            if (context.GetFeature<IStatusCodePagesFeature>() == null)
+            {
+                // No-op if the feature is turned off or missing.
```
Eilon: Recommended new comment :

    // Check if the feature is still available because other middleware (such as a web API written in MVC) could
    // have removed the feature to prevent HTML status code responses from showing up to an API client.

src/Microsoft.AspNet.Diagnostics/IStatusCodePagesFeature.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.Runtime;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    [AssemblyNeutral]
+    public interface IStatusCodePagesFeature
```
Eilon: Actually a doc comment on this one is more important.

Tratcher:
I can't think of any other feature interfaces without members. The Marker-Interface pattern is something we try to avoid. I'd rather have the behavior determined by the property than the presence of the interface.

Praburaj:
hmm. Interesting.. okay. I can then place the `Enabled` property back.

src/Microsoft.AspNet.Diagnostics/project.json
```diff
@@ -2,16 +2,16 @@
     "version": "1.0.0-*",
     "description": "ASP.NET 5 Middleware for error handling, error pages, and diagnostics information.",
     "dependencies": {
+        "Microsoft.AspNet.Diagnostics.Interfaces": "1.0.0-*",
```
Praburaj: I guess this can become a build type dependency.

Praburaj:
@yishaigalatzer @Tratcher updated PR.

src/Microsoft.AspNet.Diagnostics/project.json
Eilon: I think the build type dependency would work only with ANIs. This is now a real dependency on a real type that is really linked to, no?

Praburaj: hmm. okay.

Eilon: You don't seem convinced :smile: If the type is a regular type (*not* an ANI), why *wouldn't* a regular reference be needed? Where would the type come from, if not from that package?

Praburaj: oh no convinced :smile:

Eilon: I came home early today because I wasn't feeling well, so feel free to be unconvinced if you think I'm mistaken :smile: (And then I'll just blame my not feeling well!)

Praburaj: oh no. I would let you rest :smile:

src/Microsoft.AspNet.Diagnostics.Elm/project.json
```diff
@@ -2,8 +2,7 @@
     "version": "1.0.0-*",
     "dependencies": {
         "Microsoft.AspNet.Diagnostics": "1.0.0-*",
-        "Microsoft.Framework.Logging.Interfaces": { "version": "1.0.0-*", "type": "build" },
-        "Microsoft.Framework.Runtime.Interfaces": { "version": "1.0.0-*", "type": "build" }
+        "Microsoft.Framework.Logging.Interfaces": { "version": "1.0.0-*", "type": "build" }
```
Tratcher: No more build time dependencies.

Tratcher:
:shipit:

Issue #89, opened 2015-02-13
Prefer the new encoders over the old WebUtility APIs


GrabYourPitchforks:
There are still some call sites in the Diagnostics code base that invoke WebUtility.HtmlEncode.  These call sites should instead use IHtmlEncoder.HtmlEncoder.

GrabYourPitchforks:
The types `IHtmlEncoder`, `IJavaScriptStringEncoder`, and `IUrlEncoder` are in the Microsoft.AspNet.WebUtilities package. To register them from within your own AddServices method, reference the Microsoft.AspNet.Http.Extensions package and call `AddEncoders`.

muratg:
Looks like the PR is pending?  @Praburaj

Praburaj:
Yes. It will be fixed soon as I found an issue with the `GetHtmlEncoder` extension.

PR #90, opened 2015-02-17
Using IHtmlEncoder to encode content

Praburaj:
Fixes: https://github.com/aspnet/Diagnostics/issues/89

@GrabYourPitchforks @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
+++ b/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
@@ -257,7 +257,6 @@ namespace Microsoft.AspNet.Diagnostics.Views
             }
 
             // Split on line breaks before passing it through the encoder.
-            // We use the static default encoder since we can't depend on DI in the error handling logic.
             return string.Join("<br />" + Environment.NewLine,
                 input.Split(new[] { "\r\n" }, StringSplitOptions.None)
                 .SelectMany(s => s.Split(new[] { '\r', '\n' }, StringSplitOptions.None))
--- a/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
+++ b/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
@@ -5,9 +5,10 @@
 using System;
 using System.Globalization;
 using System.IO;
-using System.Net;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
+using Microsoft.Framework.WebEncoders;
 
 namespace Microsoft.AspNet.Diagnostics.Views
 {
@@ -37,6 +38,11 @@ namespace Microsoft.AspNet.Diagnostics.Views
         protected StreamWriter Output { get; private set; }
 
         /// <summary>
+        /// Html encoder used to encode content.
+        /// </summary>
+        protected IHtmlEncoder HtmlEncoder { get; set; }
+
+        /// <summary>
         /// Execute an individual request
         /// </summary>
         /// <param name="context"></param>
@@ -46,6 +52,7 @@ namespace Microsoft.AspNet.Diagnostics.Views
             Request = Context.Request;
             Response = Context.Response;
             Output = new StreamWriter(Response.Body);
+            HtmlEncoder = context.ApplicationServices.GetHtmlEncoder();
             await ExecuteAsync();
             Output.Dispose();
         }
@@ -217,7 +224,7 @@ namespace Microsoft.AspNet.Diagnostics.Views
         /// <param name="value">The <see cref="string"/> to write.</param>
         protected void WriteTo(TextWriter writer, string value)
         {
-            WriteLiteralTo(writer, WebUtility.HtmlEncode(value));
+            WriteLiteralTo(writer, HtmlEncoder.HtmlEncode(value));
         }
 
         /// <summary>
@@ -240,6 +247,21 @@ namespace Microsoft.AspNet.Diagnostics.Views
             {
                 writer.Write(value);
             }
+        }
+
+        protected string HtmlEncodeAndReplaceLineBreaks(string input)
+        {
+            if (string.IsNullOrEmpty(input))
+            {
+                return string.Empty;
+            }
+
+            // Split on line breaks before passing it through the encoder.
+            // We use the static default encoder since we can't depend on DI in the error handling logic.
+            return string.Join("<br />" + Environment.NewLine,
+                input.Split(new[] { "\r\n" }, StringSplitOptions.None)
+                .SelectMany(s => s.Split(new[] { '\r', '\n' }, StringSplitOptions.None))
+                .Select(HtmlEncoder.HtmlEncode));
         }
     }
 }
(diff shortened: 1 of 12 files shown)


GrabYourPitchforks:
:shipit:, with one comment. Consider using the `IServiceProvider.GetHtmlEncoder()` extension method instead of calling `GetRequiredService`. It will give you a default encoder instance if one hasn't already been registered in the service provider. One benefit of this is that it doesn't punish developers who forget to call AddDiagnostics.

GrabYourPitchforks:
One other comment: the namespace `Microsoft.AspNet.WebUtilities.Encoders` has been renamed to `Microsoft.Framework.WebEncoders`. You'll need to update before you merge.

samples/ErrorPageSample/Startup.cs
```diff
@@ -1,12 +1,20 @@
...
     public class Startup
     {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddEncoders();
+        }
+
         public void Configure(IApplicationBuilder app)
         {
+            app.UseRequestServices();
```
davidfowl: This isn't required right?

davidfowl:
I'm not so sure about this change. Adding the encoder shouldn't be required by these packages. The razor pages are used as an implementation detail and the app code shoudn't have to provide an encoder for the error page to work. 

We should use the static version in the razor pages in these packages.

GrabYourPitchforks:
The general idea is that if you have an instance of an `IServiceProvider`, you can call the `GetHtmlEncoder()` extension method, and it is guaranteed to return non-null. (It returns `HtmlEncoder.Default` if nothing is available in the service provider.) If you don't have access to an `IServiceProvider` instance, just use the `HtmlEncoder.Default` static singleton. The API should be easy to consume both in the DI case and in the non-DI case; there's no need for contortions.

samples/ErrorPageSample/Startup.cs
Praburaj: Nope. We now have a default `IHtmlEncoder`. I have to update this PR with that change.

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cs
```diff
@@ -95,7 +95,7 @@ public CompilationErrorPage()
 #line hidden
             WriteLiteral(": ");
 #line 30 "CompilationErrorPage.cshtml"
-                                                                          Output.Write(WebUtility.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>")); 
+                                                                          Output.Write(HtmlEncoder.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>"));
```
GrabYourPitchforks: Line breaks must be stripped before passing the string through the encoder, as the encoder could replace the line breaks with whatever representation it wishes. See https://github.com/aspnet/Helios/blob/dev/src/Microsoft.AspNet.Loader.IIS/Util/ErrorHandlingExtensions.cs#L350 for an example.

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cshtml
```diff
@@ -27,7 +32,7 @@
         <h1>@Resources.ErrorPageHtml_CompilationException</h1>
         @if (Model.Options.ShowExceptionDetails)
         {
-            <div class="titleerror">@errorDetail.Error.GetType().Name: @{ Output.Write(WebUtility.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>")); }</div>
+            <div class="titleerror">@errorDetail.Error.GetType().Name: @{ Output.Write(_htmlEncoder.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>")); }</div>
```
GrabYourPitchforks: Same comment as above - perform the line break check separately.

src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cs
```diff
@@ -105,7 +105,7 @@ public ErrorPage()
 #line hidden
             WriteLiteral(": ");
 #line 37 "ErrorPage.cshtml"
-                                                                              Output.Write(WebUtility.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>")); 
+                                                                              Output.Write(HtmlEncoder.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>"));
```
GrabYourPitchforks: Line break check.

src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cshtml
```diff
@@ -34,7 +38,7 @@
         {
             foreach (var errorDetail in Model.ErrorDetails)
             {
-                <div class="titleerror">@errorDetail.Error.GetType().Name: @{ Output.Write(WebUtility.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>")); }</div>
+                <div class="titleerror">@errorDetail.Error.GetType().Name: @{ Output.Write(_htmlEncoder.HtmlEncode(errorDetail.Error.Message).Replace("\r\n", "<br/>").Replace("\r", "<br/>").Replace("\n", "<br/>")); }</div>
```
GrabYourPitchforks: Line break check.

src/Microsoft.AspNet.Diagnostics/project.json
```diff
@@ -3,6 +3,7 @@
     "description": "ASP.NET 5 Middleware for error handling, error pages, and diagnostics information.",
     "dependencies": {
         "Microsoft.AspNet.Diagnostics.Interfaces": "1.0.0-*",
+        "Microsoft.AspNet.Http.Extensions": "1.0.0-*",
```
GrabYourPitchforks: Don't need `Microsoft.AspNet.Http.Extensions` any more since the extension method have been moved out. Use `Microsoft.Framework.WebEncoders` directly instead.

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
```diff
@@ -46,6 +52,7 @@ public abstract class BaseView
             Request = Context.Request;
             Response = Context.Response;
             Output = new StreamWriter(Response.Body);
+            HtmlEncoder = context.ApplicationServices.GetRequiredService<IHtmlEncoder>();
```
GrabYourPitchforks: Consider the ApplicationServices.GetHtmlEncoder() extension method, as it will return a default instance if one hasn't been registered with DI.

GrabYourPitchforks:
:watch:

Praburaj:
@GrabYourPitchforks updated PR.

GrabYourPitchforks:
:shipit:

Issue #96, opened 2015-03-12
IServiceCollection extension for Elm should live in Microsoft.Framework.DependencyInjection namespace like all other extensions


Praburaj:
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics.Elm/ElmExtensions.cs#L40

PR #97, opened 2015-03-12
Move IServiceCollection extension to Microsoft.Framework.DependencyInjec...

Praburaj:
...tion namespace

Fixes: https://github.com/aspnet/Diagnostics/issues/96

@muratg

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics.Elm/ElmServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Diagnostics.Elm/ElmServiceCollectionExtensions.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Jetbrains.Annotations;
+using Microsoft.AspNet.Diagnostics.Elm;
+
+namespace Microsoft.Framework.DependencyInjection
+{
+    public static class ElmServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Registers an <see cref="ElmStore"/> and configures <see cref="ElmOptions"/>.
+        /// </summary>
+        public static IServiceCollection AddElm([NotNull] this IServiceCollection services)
+        {
+            return services.AddElm(configureOptions: null);
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Diagnostics.Elm/ElmServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Jetbrains.Annotations;
+using Microsoft.Framework.DependencyInjection;
+
+namespace Microsoft.AspNet.Diagnostics.Elm
```
Praburaj: Oops. Fixing the namespace :smile:

src/Microsoft.AspNet.Diagnostics.Elm/ElmServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.AspNet.Diagnostics.Elm;
+
+namespace Microsoft.Framework.DependencyInjection
+{
+    public static class ElmServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Registers an <see cref="ElmStore"/> and configures <see cref="ElmOptions"/>.
+        /// </summary>
+        public static IServiceCollection AddElm([NotNull] this IServiceCollection services, Action<ElmOptions> configureOptions = null)
```
davidfowl: No more options params.

Praburaj: you mean optional params?

Praburaj:
Updated removing optional parameters.

src/Microsoft.AspNet.Diagnostics.Elm/ElmServiceCollectionExtensions.cs
muratg: Yep.  Default values are ok only for CancellationTokens. Remove ```= null``` from the end and pass it explicitly where needed.

muratg:
Looks good, unless @davidfowl has more comments, :shipit:

PR #99, opened 2015-03-14
using [NotNull] attribute from the Common repo

Praburaj:
@muratg


src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
```diff
@@ -2,9 +2,10 @@
 #if DNX451
 using System.Runtime.Remoting;
 using System.Runtime.Remoting.Messaging;
-#endif
+#else
 using System.Threading;
```
muratg: System.Threading seems to be always included before this change.  Not it's not included for DNX451.  Why this change?

Praburaj: This was used only in the #else case. So VS intellisense is always showing this is an unused using statement. So moved it into the #else to clean up the intellisense warning.

Praburaj: and of course I verified project builds.

muratg:
:shipit:

PR #106, opened 2015-03-24
Fixing the Elm middleware to check if request has form content

Praburaj:
With a change to HttpAbstractions, ReadFormAsync throws if called on a request whose content type is not
not form encoded content.

@muratg


muratg:
:shipit:

PR #113, opened 2015-03-31
Fix ElmScope to consider format string when creating scope message.

kichalla:
@suhasj @rynowak @loudej 

Related(but not dependent) PR: https://github.com/aspnet/Logging/pull/159

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
+++ b/src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
@@ -6,6 +6,7 @@ using System.Runtime.Remoting.Messaging;
 using System.Threading;
 #endif
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Diagnostics.Elm
 {
@@ -67,10 +68,22 @@ namespace Microsoft.AspNet.Diagnostics.Elm
             Current = scope;
             Current.Parent = temp;
 
+            // Get the scope message
+            string scopeMessage = null;
+            var scopeLogValues = Current._state as ILogValues;
+            if(scopeLogValues != null)
+            {
+                scopeMessage = scopeLogValues.Format();
+            }
+            else
+            {
+                scopeMessage = Current._state.ToString();
+            }
+
             Current.Node = new ScopeNode()
             {
                 StartTime = DateTimeOffset.UtcNow,
-                State = Current._state,
+                State = scopeMessage,
                 Name = Current._name
             };
 
--- a/src/Microsoft.AspNet.Diagnostics.Elm/ScopeNode.cs
+++ b/src/Microsoft.AspNet.Diagnostics.Elm/ScopeNode.cs
@@ -15,7 +15,7 @@ namespace Microsoft.AspNet.Diagnostics.Elm
 
         public DateTimeOffset EndTime { get; set; }
 
-        public object State { get; set; }
+        public string State { get; set; }
 
         public string Name { get; set; }
 


msotclas:
Hi __@kichalla__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Kiran Challa). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
```diff
@@ -67,10 +68,22 @@ public static IDisposable Push([NotNull] ElmScope scope, [NotNull] ElmStore stor
             Current = scope;
             Current.Parent = temp;
 
+            // Get the scope message
+            string scopeMessage = null;
+            var scopeLogValues = Current._state as ILogValues;
+            if(scopeLogValues != null)
+            {
+                scopeMessage = scopeLogValues.Format();
```
rynowak: /cc @loudej I think the plan here to get rid of `.Format()` and just have `.ToString()` do the right thing. Should we do that change instead?

src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
```diff
@@ -67,10 +68,22 @@ public static IDisposable Push([NotNull] ElmScope scope, [NotNull] ElmStore stor
...
+            else
+            {
+                scopeMessage = Current._state.ToString();
+            }
+
             Current.Node = new ScopeNode()
             {
                 StartTime = DateTimeOffset.UtcNow,
-                State = Current._state,
+                State = scopeMessage,
```
rynowak: I'm not familiar enough with ELM to really know if this is right.

src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
```diff
@@ -67,10 +68,22 @@ public static IDisposable Push([NotNull] ElmScope scope, [NotNull] ElmStore stor
             Current = scope;
             Current.Parent = temp;
 
+            // Get the scope message
+            string scopeMessage = null;
+            var scopeLogValues = Current._state as ILogValues;
+            if(scopeLogValues != null)
```
Eilon: Formatting :lipstick:

src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
kichalla: @loudej @lodejard : Should I get rid of `.Format()` all together? I ask this because as part of the rename of `BeginScope` to `BeginScopeImpl`, I am coming across code (across repos) where `ToString()`  is being invoked on the state. Depending on what we decide here it would impact the number of changes that I will have to make across the repos to reflect the renaming change.

lodejard: Yep, no need for format

kichalla: Ah, noticed this just now before I sent out the other PR in logging repo

kichalla:
Closing this PR as these changes are no longer required due to following changes in logging recently:
https://github.com/aspnet/Logging/commit/d2ff516e8d01b867abf04ed7c41d58cb35f72a8a

PR #119, opened 2015-04-06
Using in memory store on mono

Praburaj:
@bricelam


Praburaj:
@davidfowl - The test went past the the localdb issue and failing with NullReference on `ServiceProviderExtensions.GetRequiredService` now.

Issue #118, opened 2015-04-05
Remove Hosting dependency 


davidfowl:
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/project.json#L6

Figure out what's minimally required

PR #122, opened 2015-04-08
Remove hosting dependency on diagnostics package

Praburaj:
Fixes: #118 

@davidfowl


Praburaj:
@muratg

muratg:
Looks like Travis and AppVeyor are failing, any idea why?  Other than that, change looks good.

:shipit:

Praburaj:
Thanks @muratg I think I sent the PR when there was a change in EF that was not propagated by a successful Coherence-Signed build. So the build on travis ci was failing since it builds against the vnext myget feed. I hope this should succeed if Coherence-Signed has passed. But I'll double check before pushing the change.

Praburaj:
Looks like some tests are failing due to a known DI issue now. This should be independent of this change.

PR #126, opened 2015-04-16
Fixing runtime info middleware on coreclr

Praburaj:
Also adding more information like OS, runtime type and architecture to the info page.

@victorhurdugaci


Praburaj:
This is the updated runtime info page: 

image

victorhurdugaci:
Can we please make the fields one liners? So instead of 8 lines you only have 4

Praburaj:
@victorhurdugaci updated PR. Here is the new screenshot.
image

victorhurdugaci:
:shipit:

PR #137, opened 2015-05-14
React to EF: changes in ModelSnapshot

smitpatel:
Changes to match with aspnet/EntityFramework@252534cee168e257479786f94bf7b72b5c84f001
@bricelam


bricelam:
:shipit: Thanks for fixing my build break. :grin:

Issue #130, opened 2015-05-06
UseErrorPage extension rationalization


muratg:
Currently, templates check the environment, and if it's Development, UseErrorPage extension is called.

However, there's a TODO comment in this extension method: https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ErrorPageExtensions.cs#L43

```
        public static IApplicationBuilder UseErrorPage(this IApplicationBuilder builder, ErrorPageOptions options)
        {
            if (builder == null)
            {
                throw new ArgumentNullException("builder");
            }
            /* TODO: Development, Staging, or Production
            string appMode = new AppProperties(builder.Properties).Get<string>(Constants.HostAppMode);
            bool isDevMode = string.Equals(Constants.DevMode, appMode, StringComparison.Ordinal);*/
            bool isDevMode = true;
            return builder.Use(next => new ErrorPageMiddleware(next, options, isDevMode).Invoke);
        }
```

We either need to remove this comment from the method, or enable the check here, and remove UseErrorPage() in the Startup outside of Development env check.

muratg:
We had a chat with @DamianEdwards and @Eilon and decided to keep this as is.  If customers call UseErrorPage, we are not going to stop them from here.

Thus the work is removing the comments above.

Eilon:
Does the `ErrorPageMiddleware` ctor still need an `isDevMode` parameter? Sounds like we're basically saying you shouldn't really use that anyway. And if you do, you can pass in an explicit `ErrorPageOptions` with `options.SetDefaultVisibility(isVisible: true);` yourself. Though, should we even get rid of *that* option?

muratg:
Yup, I think we should remove the `isDevMode` parameter in the constructor as well as `isVisible` option in `ErrorPageOptions`.  With a cursory look, we are not even using `isVisible` option anywhere but tests.

henkmollema:
@muratg @Tratcher this can be closed I guess.

muratg:
Yup, thanks @henkmollema!

PR #140, opened 2015-06-01
Removed SetDefaultVisibility for ErrorPageOptions

henkmollema:
- Removed `SetDefaultVisibility` method from `(Database)ErrorPageOptions`
- `(Database)ErrorPageOptions` properties are now auto-properties since there is no default visibility anymore
- Removed `isDevMode` parameter from `ErrorPageMiddleware` ctor
- Removed comments from `ErroPageExtensions.UseErrorPage()`
- Updated Entity tests

#130


Tratcher:
Looks good.

Tratcher:
Merged

davidfowl:
We should remove ErrorPageOptions all together

henkmollema:
@davidfowl how will that look like? Just show everything by default?

davidfowl:
Yes, thats why the project template has a switch.

PR #149, opened 2015-07-13
[Fixes #127] ElmCaptureMiddleware should not create logging request scope

### LINKED ISSUE #127: ElmCaptureMiddleware should not create logging request scope
Since we now create request scope at hosting layer, the following should no longer be required.

https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs#L31

kichalla:
@Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs
+++ b/src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs
@@ -1,10 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.OptionsModel;
 
@@ -28,18 +28,20 @@ namespace Microsoft.AspNet.Diagnostics.Elm
 
         public async Task Invoke(HttpContext context)
         {
-            var requestId = Guid.NewGuid();
-            using (_logger.BeginScope(string.Format("request {0}", requestId)))
+            using (RequestIdentifier.Ensure(context))
             {
-                var p = ElmScope.Current;
-                ElmScope.Current.Context.HttpInfo = GetHttpInfo(context, requestId);
-                try
+                var requestId = context.GetFeature<IHttpRequestIdentifierFeature>().TraceIdentifier;
+                using (_logger.BeginScope("Request: {RequestId}", requestId))
                 {
-                    await _next(context);
-                }
-                finally
-                {
-                    ElmScope.Current.Context.HttpInfo.StatusCode = context.Response.StatusCode;
+                    try
+                    {
+                        ElmScope.Current.Context.HttpInfo = GetHttpInfo(context);
+                        await _next(context);
+                    }
+                    finally
+                    {
+                        ElmScope.Current.Context.HttpInfo.StatusCode = context.Response.StatusCode;
+                    }
                 }
             }
         }
@@ -48,11 +50,11 @@ namespace Microsoft.AspNet.Diagnostics.Elm
         /// Takes the info from the given HttpContext and copies it to an HttpInfo object
         /// </summary>
         /// <returns>The HttpInfo for the current elm context</returns>
-        private static HttpInfo GetHttpInfo(HttpContext context, Guid requestId)
+        private static HttpInfo GetHttpInfo(HttpContext context)
         {
             return new HttpInfo()
             {
-                RequestID = requestId,
+                RequestID = context.GetFeature<IHttpRequestIdentifierFeature>().TraceIdentifier,
                 Host = context.Request.Host,
                 ContentType = context.Request.ContentType,
                 Path = context.Request.Path,
--- a/src/Microsoft.AspNet.Diagnostics.Elm/RequestIdentifier.cs
+++ b/src/Microsoft.AspNet.Diagnostics.Elm/RequestIdentifier.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Diagnostics.Elm
+{
+    internal class RequestIdentifier : IDisposable
+    {
+        private readonly bool _addedFeature;
+        private readonly bool _updatedIdentifier;
+        private readonly string _originalIdentifierValue;
+        private readonly HttpContext _context;
+        private readonly IHttpRequestIdentifierFeature _feature;
+
+        private RequestIdentifier(HttpContext context)
+        {
+            _context = context;
+            _feature = context.GetFeature<IHttpRequestIdentifierFeature>();
+
+            if (_feature == null)
+            {
+                _feature = new HttpRequestIdentifierFeature()
+                {
+                    TraceIdentifier = Guid.NewGuid().ToString()
+                };
+                context.SetFeature(_feature);
+                _addedFeature = true;
+            }
+            else if (string.IsNullOrEmpty(_feature.TraceIdentifier))
+            {
+                _originalIdentifierValue = _feature.TraceIdentifier;
+                _feature.TraceIdentifier = Guid.NewGuid().ToString();
+                _updatedIdentifier = true;
+            }
+        }
+
+        public static IDisposable Ensure(HttpContext context)
+        {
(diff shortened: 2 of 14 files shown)


src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs
```diff
@@ -48,11 +45,11 @@ public ElmCaptureMiddleware(RequestDelegate next, ILoggerFactory factory, IOptio
         /// Takes the info from the given HttpContext and copies it to an HttpInfo object
         /// </summary>
         /// <returns>The HttpInfo for the current elm context</returns>
-        private static HttpInfo GetHttpInfo(HttpContext context, Guid requestId)
+        private static HttpInfo GetHttpInfo(HttpContext context)
         {
             return new HttpInfo()
             {
-                RequestID = requestId,
+                RequestID = context.GetFeature<IHttpRequestIdentifierFeature>().TraceIdentifier,
```
Tratcher: IHttpRequestIdentifierFeature may be null.

kichalla: Actually we  create a default one if it is not present https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs#L162

Tratcher: GetFeature can never assume anything about other components. Hosting may add it but other middleware can modify or remove it.

kichalla: Good point, in that case probably I should have fallback behavior to create a request scope id as the current code was doing?

Tratcher: You can't tell if you're inside a logging scope, correct? I wouldn't conditionally create one. Do or don't.

Conditionally creating a TraceIdentifier would be good, but if you do that you should assign it back to the request.

kichalla: Sure, updated now. cc: @Tratcher

src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs
```diff
@@ -48,11 +45,11 @@ public ElmCaptureMiddleware(RequestDelegate next, ILoggerFactory factory, IOptio
         /// Takes the info from the given HttpContext and copies it to an HttpInfo object
         /// </summary>
         /// <returns>The HttpInfo for the current elm context</returns>
-        private static HttpInfo GetHttpInfo(HttpContext context, Guid requestId)
+        private static HttpInfo GetHttpInfo(HttpContext context)
         {
             return new HttpInfo()
             {
-                RequestID = requestId,
+                RequestID = context.GetFeature<IHttpRequestIdentifierFeature>()?.TraceIdentifier,
```
Tratcher: If it's missing create a default and set it back on the context.

src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs
```diff
@@ -28,19 +29,15 @@ public ElmCaptureMiddleware(RequestDelegate next, ILoggerFactory factory, IOptio
 
         public async Task Invoke(HttpContext context)
         {
-            var requestId = Guid.NewGuid();
-            using (_logger.BeginScope(string.Format("request {0}", requestId)))
+            var p = ElmScope.Current;
```
Tratcher: what's the point of this line? It looks like a no-op. It doesn't look like it did anything before either.

In fact, it looks like the next line would easily null ref.

kichalla:
Updated. @Tratcher

src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs
```diff
@@ -28,18 +28,20 @@ public ElmCaptureMiddleware(RequestDelegate next, ILoggerFactory factory, IOptio
...
+                using (_logger.BeginScope("Request: {RequestId}", requestId))
                 {
-                    await _next(context);
-                }
-                finally
-                {
-                    ElmScope.Current.Context.HttpInfo.StatusCode = context.Response.StatusCode;
+                    try
+                    {
+                        ElmScope.Current.Context.HttpInfo = GetHttpInfo(context);
```
Tratcher: The only part I can't figure out is where ElmScope.Current is initialized. It looks like it would just null ref.

src/Microsoft.AspNet.Diagnostics.Elm/RequestIdentifier.cs
```diff
@@ -0,0 +1,42 @@
...
+            _context = context;
+            var feature = context.GetFeature<IHttpRequestIdentifierFeature>();
+            if (feature == null)
+            {
+                context.SetFeature<IHttpRequestIdentifierFeature>(new HttpRequestIdentifierFeature()
+                {
+                    TraceIdentifier = Guid.NewGuid().ToString()
+                });
+                _addedFeature = true;
+            }
```
Tratcher: what if feature != null but string.IsNullOrEmpty(feature.TraceIdentifier)?

src/Microsoft.AspNet.Diagnostics.Elm/ElmCaptureMiddleware.cs
kichalla: When a `_logger.BeginScope` is called https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics.Elm/ElmLogger.cs#L62
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs#L67

src/Microsoft.AspNet.Diagnostics.Elm/RequestIdentifier.cs
kichalla: Ok...sure, will update...

src/Microsoft.AspNet.Diagnostics.Elm/Views/LogPage.cs
```diff
@@ -306,6 +306,8 @@ public LogPage()
         #pragma warning disable 1998
         public override async Task ExecuteAsync()
         {
+            Response.ContentType = "text/html; charset=utf-8";
+            Response.ContentLength = null; // Clear any prior Content-Length
```
Tratcher: There shouldn't be a prior content-length for either of these pages, did you see one?

kichalla: I followed what other view pages were doing : https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cs#L60-L61

Tratcher: Those are calling _next and then catching exceptions. This middleware does not call _next if it finds a path match.

kichalla: Ah ok...I have now removed this change and updated the PR..

src/Microsoft.AspNet.Diagnostics.Elm/RequestIdentifier.cs
```diff
@@ -0,0 +1,42 @@
...
+    internal class RequestIdentifier : IDisposable
+    {
+        private readonly bool _addedFeature;
+        private readonly HttpContext _context;
+
+        private RequestIdentifier(HttpContext context)
+        {
+            _context = context;
+            var feature = context.GetFeature<IHttpRequestIdentifierFeature>();
+            if (feature == null || (feature != null && string.IsNullOrEmpty(feature.TraceIdentifier)))
```
Tratcher: The feature != null check here is redundant.

Tratcher: If the feature exists but the TraceIdentifier doesn't, just set the TraceIdentifier and not _addedFeature.

kichalla: Ok, but is this even a possible scenario that we would be seeing?

Tratcher: Possible? yes. Likely? probably not.If this type ways trying to ensure the Feature then I wouldn't worry about it, but you're trying to ensure the value, not the feature.

kichalla: @Tratcher Sure, I made the changes and added tests too. Please take a look.

Tratcher:
:shipit:

kichalla:
commit fea15eb8de4bc36debe32e788e8ab5868ea374df

PR #151, opened 2015-07-13
[Fixes #112] BaseView.ExecuteAsync disposes the response stream

### LINKED ISSUE #112: BaseView.ExecuteAsync disposes the response stream
I was looking at [this line](https://github.com/aspnet/Diagnostics/blob/9cffbd171b3f1fdd61de251685984c885e5efadb/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs#L58) and I got confused a little. It disposes the stream which is basically the response stream. 

Should this be flush instead of dispose?

kichalla:
@Tratcher


src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
```diff
@@ -55,7 +55,7 @@ public abstract class BaseView
             Output = new StreamWriter(Response.Body);
```
kichalla: I tried the `AutoFlush=true` but didn't work as its supposed to probably because of the following?(BTW I tried this with `ErrorPageSample` in the solution):

https://msdn.microsoft.com/en-us/library/system.io.streamwriter.autoflush(v=vs.110).aspx

"Flushing the stream will not flush its underlying encoder unless you explicitly call Flush or Close. Setting AutoFlush to true means that data will be flushed from the buffer to the stream after each write operation, but the encoder state will not be flushed. This allows the encoder to keep its state (partial characters) so that it can encode the next block of characters correctly. This scenario affects UTF8 and UTF7 where certain characters can only be encoded after the encoder receives the adjacent character or characters."

Tratcher: How about using the StreamWriter constructor with the leaveOpen parameter? Then you can dispose the StreamWriter.

kichalla: Sure, sounds good.

kichalla: Updated. cc: @Tratcher

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
```diff
@@ -52,7 +53,7 @@ public abstract class BaseView
             Context = context;
             Request = Context.Request;
             Response = Context.Response;
-            Output = new StreamWriter(Response.Body);
+            Output = new StreamWriter(Response.Body, Encoding.UTF8, 1024, leaveOpen: true);
```
Tratcher: hmm, we don't set the charset in the content-type header do we? Can you add that while you're here?

Also, raise the buffer to 4kb. These pages are pretty big.

kichalla: The latest commit has charset set and increased buffer size. updated. @Tratcher

Tratcher:
:shipit:

PR #152, opened 2015-07-13
[Fixes #40] Suppress caching for ErrorHandler

### LINKED ISSUE #40: Suppress caching for ErrorHandler
When using the ErrorHandler to re-execute a failed request and generate a page, that error page should not be cached. However, components like the static file middleware cache by default via e-tags.

Consider modifying the samples to manually add no-cache headers on the app-error path, or modify the middleware itself to set these no-cache headers. before re-executing the request.

kichalla:
@Tratcher


src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs
```diff
@@ -70,9 +74,17 @@ public ErrorHandlerMiddleware(RequestDelegate next, ILoggerFactory loggerFactory
...
                 throw; // Re-throw the original if we couldn't handle it
             }
         }
+
+        private Task ClearCacheHeaders(object state)
+        {
+            var response = (HttpResponse)state;
+            response.Headers.Set("Cache-Control", "no-cache");
+            response.Headers.Set("Pragma", "no-cache");
+            response.Headers.Set("Expires", "-1");
```
kichalla: I suppose I should not be doing anything regarding ETag headers?

Tratcher: I'd recommend removing the e-tag header as well.

kichalla: Updated. cc: @Tratcher

Tratcher:
Add a few tests and :shipit:

kichalla:
@Tratcher added tests for this. can you take a look? (especially about the new project.json references)

test/Microsoft.AspNet.Diagnostics.Tests/project.json
```diff
@@ -4,6 +4,9 @@
     },
     "dependencies": {
         "Microsoft.AspNet.Diagnostics.Elm": "1.0.0-*",
+        "Microsoft.AspNet.FileProviders.Embedded": "1.0.0-*",
+        "Microsoft.AspNet.StaticFiles": "1.0.0-*",
```
Tratcher: Middleware repos should avoid taking dependencies on other middleware repos. It complicates the CI dependency chain. You should be able to mimic the behavior of StaticFiles accurately enough for testing.

kichalla: Sure, I have removed the test specifically targeting StaticFiles as this scenario is already being covered in the other test that I added. Also added new tests. @Tratcher

test/Microsoft.AspNet.Diagnostics.Tests/ErrorHandlerTest.cs
```diff
@@ -0,0 +1,257 @@
...
+
+                app.Run(async (httpContext) =>
+                {
+                    await httpContext.Response.WriteAsync("Hello");
+                    throw new InvalidOperationException("Something bad happened");
+                });
+            }))
+            {
+                var client = server.CreateClient();
+                var response = await client.GetAsync(string.Empty);
```
Tratcher: Doesn't the TestServer throw the InvalidOperationException back to you here?

Tratcher:
squash and :shipit:

PR #155, opened 2015-07-14
Fix argument null exception when rendering an error page

kichalla:
@NTaylorMullen @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cs
+++ b/src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cs
@@ -193,7 +193,7 @@ using Views
 #line hidden
 
 #line 48 "CompilationErrorPage.cshtml"
-                         if (frame.Line != 0 && frame.ContextCode.Any())
+                         if (frame.Line != 0 && frame.ContextCode !=null && frame.ContextCode.Any())
                         {
 
 #line default
--- a/src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cshtml
+++ b/src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cshtml
@@ -45,7 +45,7 @@
                             <h3>@frame.ErrorDetails</h3>
                         }
 
-                        @if (frame.Line != 0 && frame.ContextCode.Any())
+                        @if (frame.Line != 0 && frame.ContextCode !=null && frame.ContextCode.Any())
                         {
                             <div class="source">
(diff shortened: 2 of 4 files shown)


NTaylorMullen:
Question: Why was `frameContext` not being set?

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cs
```diff
@@ -193,7 +193,7 @@ public CompilationErrorPage()
 #line hidden
 
 #line 48 "CompilationErrorPage.cshtml"
-                         if (frame.Line != 0 && frame.ContextCode.Any())
+                         if (frame.Line != 0 && frame.ContextCode !=null && frame.ContextCode.Any())
```
NTaylorMullen: `frame.ContextCode?.Any() == true` for this and below

kichalla:
@NTaylorMullen I am not exactly sure but we are populating this data for ContextCode
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs#L202

For this particular case it looks like below:
image

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cs
kichalla: I think the one I have is probably more readable?

NTaylorMullen: Maybe it's just a repo preference, in Razor we do the `?. == true` trick a lot. Not vastly important.

NTaylorMullen:
@kichalla Should definitely track down why it's being set to `null` (why it's failing to locate the responsible code). The original code has `null` checks for pre/post context code which makes me believe it was purposefully left without the normal context code `null` checks. 

Want to make sure we're not solving the wrong problem here.

kichalla:
So on digging deeper I found that the real cause of this issue is with ErrorPage middleware not showing source code lines of Razor files. I filed a new issue here: https://github.com/aspnet/Diagnostics/issues/157

However the fix in this PR is still good as we want to avoid throwing argument null exception.

NTaylorMullen:
Got it, so once #157 is completed the only way an error could occur is if a user put in a faulty `#line` pragma in their application (which would result in a `null` `frame.ContextCode`).

:shipit:

kichalla:
commit 191ffe17219405c6c7c9607c20ba6dd6fbdeecde

PR #156, opened 2015-07-14
Added logging to ErrorPageMiddleware

kichalla:
@Tratcher


src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -51,14 +55,16 @@ public ErrorPageMiddleware([NotNull] RequestDelegate next, [NotNull] ErrorPageOp
             }
             catch (Exception ex)
             {
+                _logger.LogError("An unhandled exception has occurred while executing the request", ex);
```
kichalla: I noticed that `ErrorHandler` middleware already has some logging...not sure if we are looking for more.
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs#L37

src/Microsoft.AspNet.Diagnostics/ErrorPageExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static IApplicationBuilder UseErrorPage([NotNull] this IApplicationBuilde
         /// <returns></returns>
         public static IApplicationBuilder UseErrorPage([NotNull] this IApplicationBuilder builder, ErrorPageOptions options)
         {
-            return builder.Use(next => new ErrorPageMiddleware(next, options).Invoke);
+            return builder.UseMiddleware<ErrorPageMiddleware>(options);
```
kichalla: I have tested this change where it populates the logger factory parameter, so i think this change is fine

Tratcher:
:shipit:

PR #158, opened 2015-07-17
[Fixes #157] ErrorPage middleware doesn't show source code lines of R…

### LINKED ISSUE #157: ErrorPage middleware doesn't show source code lines of Razor files
Stack trace:
image

Error page:
image

Cause(Here the File.Exists check returns false for the path `/Views/Home/Index.cshtml`):
```c#
private StackFrame LoadFrame(string function, string file, int lineNumber)
{
    var frame = new StackFrame { Function = function, File = file, Line = lineNumber };
    if (File.Exists(file))
    {
        var code = File.ReadLines(file);
        ReadFrameContent(frame, code, lineNumber, lineNumber);
    }
    return frame;
}
```

cc: @NTaylorMullen @Tratcher

kichalla:
…azor files

@NTaylorMullen @Tratcher


test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,186 @@
...
+            // Lines 4-17 (inclusive) is the code block
+            Assert.Equal(4, stackFrame.PreContextLine);
+            Assert.Equal<string>(GetCodeLines(4, 6), stackFrame.PreContextCode);
+            Assert.Equal("Line10", stackFrame.ContextCode.First());
+            Assert.Equal<string>(GetCodeLines(11, 6), stackFrame.PostContextCode);
+        }
+
+        [Theory]
+        [InlineData("TestFiles/EmbeddedSourceFile.txt")]
+        //[InlineData(@"TestFiles\EmbeddedSourceFile.txt")]
```
kichalla: @victorhurdugaci This kind of relative path works when using PhysicalFileProvider but does not when using Embedded...since in any case we would be replacing `/` with a `.`, should we even handle `\` too?

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+                {
+                    // ReadLines doesn't accept a stream. Use ReadLines as its more efficient
+                    // relative to reading lines via stream reader
+                    if (!string.IsNullOrEmpty(fileInfo.PhysicalPath))
+                    {
+                        lines = File.ReadLines(fileInfo.PhysicalPath);
+                    }
+                    else
+                    {
+                        lines = GetLines(fileInfo.CreateReadStream());
```
kichalla: The case here is that Razor files could be coming from embedded resources too and Razor view engine supports it.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+            frame.ContextCode = new[] { codeBlock[errorLineNumberInArray] };
+            frame.PostContextCode = codeBlock
+                .Skip(errorLineNumberInArray + 1)
+                .Take(codeBlock.Length - errorLineNumberInArray);
+        }
+
+        private static IEnumerable<string> GetLines(Stream stream)
         {
-            frame.PreContextLine = Math.Max(startLineNumber - _options.SourceCodeLineCount, 1);
-            frame.PreContextCode = code.Skip(frame.PreContextLine - 1).Take(startLineNumber - frame.PreContextLine).ToArray();
```
kichalla: These block of lines had to be fixed as otherwise they were closing the stream after this line

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+            // This will cause the backing stream to be closed.
+            var preErrorLineNumberInFile = Math.Max(errorLineNumberInFile - sourceCodeLineCount, 1);
+            var postErrorLineNumberInFile = errorLineNumberInFile + sourceCodeLineCount;
+            var codeBlock = allLines
+                .Skip(preErrorLineNumberInFile - 1)
+                .Take(postErrorLineNumberInFile - preErrorLineNumberInFile + 1)
+                .ToArray();
+
+            // get the position of the error line in the array
+            var errorLineNumberInArray = (errorLineNumberInFile - preErrorLineNumberInFile);
```
victorhurdugaci: nit: parenthesis not needed

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+
+            frame.PreContextLine = preErrorLineNumberInFile;
+            frame.PreContextCode = codeBlock.Take(errorLineNumberInArray);
```
victorhurdugaci: Convert these IEnumerables to list or array. Otherwise, the expressions get executed multiple times

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
victorhurdugaci: Well... the embedded file provider changes the names. It is not hierarchical

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -109,7 +115,7 @@ private Task DisplayException(HttpContext context, Exception ex)
                         Function = string.Empty
                     };
 
-                    ReadFrameContent(frame, fileContent, item.StartLine, item.EndLine);
+                    GetErrorCodeBlock(frame, fileContent, item.StartLine, item.EndLine);
```
NTaylorMullen: Q: Is `item.EndLine` really a line count?

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+            IEnumerable<string> allLines,
+            int errorLineNumberInFile,
+            int sourceCodeLineCount)
+        {
+            // Get the line boundaries in the file to be read and read all these lines at once into an array.
+            // This will cause the backing stream to be closed.
+            var preErrorLineNumberInFile = Math.Max(errorLineNumberInFile - sourceCodeLineCount, 1);
+            var postErrorLineNumberInFile = errorLineNumberInFile + sourceCodeLineCount;
+            var codeBlock = allLines
+                .Skip(preErrorLineNumberInFile - 1)
```
NTaylorMullen: is the `-1` purely for the case that `errorLineNumberInFile` > `sourceCodeLineCount`? Or is it a magic number? If it's the first, just make the `Math.Max` do `0` instead of `1`

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: Save the `errorLineNumberInFile - preErrorLineNumberInFile` value, you use it in the `Take` above.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+                .Take(postErrorLineNumberInFile - preErrorLineNumberInFile + 1)
+                .ToArray();
+
+            // get the position of the error line in the array
```
NTaylorMullen: Variable name says this.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+            var errorLineNumberInArray = (errorLineNumberInFile - preErrorLineNumberInFile);
+
+            frame.PreContextLine = preErrorLineNumberInFile;
+            frame.PreContextCode = codeBlock.Take(errorLineNumberInArray);
+            frame.ContextCode = new[] { codeBlock[errorLineNumberInArray] };
+            frame.PostContextCode = codeBlock
+                .Skip(errorLineNumberInArray + 1)
+                .Take(codeBlock.Length - errorLineNumberInArray);
```
NTaylorMullen: Why do you need to `.Take`? Shouldn't the remainder of the `codeBlock` suffice for the `PostContextCode`?

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: If `errorLineNumberInArray == 0` set `PreContextCode = null` This way you don't generate any of the HTML associated with `PreContextCode` when there isn't any. You should do the same barrier checking for `PostContextCode` when the `codeBlock` only contains `PreContextCode` and the actual line (or just the error line).

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
-                                      int startLineNumber,
-                                      int endLineNumber)
+        private static void GetErrorCodeBlock(
+            StackFrame frame,
+            IEnumerable<string> allLines,
+            int errorLineNumberInFile,
+            int sourceCodeLineCount)
+        {
+            // Get the line boundaries in the file to be read and read all these lines at once into an array.
+            // This will cause the backing stream to be closed.
```
NTaylorMullen: Where does it kill the stream in this method?

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+                    sourceCodeLineCount: _options.SourceCodeLineCount);
+            }
+            else
+            {
+                return GetStackFrame(function, file, lineNumber, _options.FileProvider, _options.SourceCodeLineCount);
+            }
         }
 
-        private StackFrame LoadFrame(string function, string file, int lineNumber)
+        internal static StackFrame GetStackFrame(
```
NTaylorMullen: Comment stating its internal for testing.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: Add a comment that `GetLines` disposes the stream indirectly or add a `using` to ensure its disposed.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+                        lines = File.ReadLines(fileInfo.PhysicalPath);
+                    }
+                    else
+                    {
+                        lines = GetLines(fileInfo.CreateReadStream());
+                    }
+                }
+            }
+
+            if (lines != null)
```
NTaylorMullen: How is it possible for lines to be `null` here?

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
 
             int lineNumber = line.ToInt32();
 
-            return string.IsNullOrEmpty(file)
-                ? LoadFrame(string.IsNullOrEmpty(function) ? line.ToString() : function, string.Empty, 0)
-                : LoadFrame(function, file, lineNumber);
+            if (string.IsNullOrEmpty(file))
+            {
+                return GetStackFrame(
+                    string.IsNullOrEmpty(function) ? line.ToString() : function,
```
NTaylorMullen: I know this was here previously but out of curiosity: Why are we passing the `line.ToString()` as the function? Doesn't that make it easy to misinterpret the responses when looking at `StackFrame`s and you see `Function` as a bit of code?

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+            }
+
             return frame;
         }
 
-        private void ReadFrameContent(StackFrame frame,
-                                      IEnumerable<string> code,
-                                      int startLineNumber,
-                                      int endLineNumber)
+        private static void GetErrorCodeBlock(
```
NTaylorMullen: `GetErrorCodeBlock` => `ReadErrorCodeBlock` or `SetErrorCodeBlock` (or something better if you can think of one). `Get` doesn't really make sense here since it's a void returning method.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: Rename to `ReadLines` to make it similar to the `File.ReadLines`

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
-            return string.IsNullOrEmpty(file)
-                ? LoadFrame(string.IsNullOrEmpty(function) ? line.ToString() : function, string.Empty, 0)
-                : LoadFrame(function, file, lineNumber);
+            if (string.IsNullOrEmpty(file))
+            {
+                return GetStackFrame(
+                    string.IsNullOrEmpty(function) ? line.ToString() : function,
+                    file: string.Empty,
+                    lineNumber: 0,
+                    fileProvider: _options.FileProvider,
```
NTaylorMullen: `_fileProvider` Make sure to add a test case for `null` file provider in options since you support it in the ctor

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,103 @@ private StackFrame StackFrame(Chunk line)
...
+                    sourceCodeLineCount: _options.SourceCodeLineCount);
+            }
+            else
+            {
+                return GetStackFrame(function, file, lineNumber, _options.FileProvider, _options.SourceCodeLineCount);
```
NTaylorMullen: ditto

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: No need for this to be static. Can just use `_fileProvider` inside it. This comment invalidates the other ones where you're using `_options.FileProvider`

src/Microsoft.AspNet.Diagnostics/ErrorPageOptions.cs
```diff
@@ -23,5 +25,7 @@ public ErrorPageOptions()
         /// source code referenced by the exception stack trace is present on the server.
         /// </summary>
         public int SourceCodeLineCount { get; set; }
+
+        public IFileProvider FileProvider { get; set; }
```
NTaylorMullen: docs

src/Microsoft.AspNet.Diagnostics/project.json
```diff
@@ -7,13 +7,17 @@
     },
     "dependencies": {
         "Microsoft.AspNet.Diagnostics.Abstractions": "1.0.0-*",
+        "Microsoft.AspNet.FileProviders.Physical": "1.0.0-*",
         "Microsoft.AspNet.Http.Extensions": "1.0.0-*",
         "Microsoft.AspNet.WebUtilities": "1.0.0-*",
+        "Microsoft.Framework.Logging.Abstractions": "1.0.0-*",
+        "Microsoft.Framework.NotNullAttribute.Sources": {
+            "type": "build",
```
NTaylorMullen: This a new format we're doing all over the place?

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.AspNet.FileProviders;
+using Xunit;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    public class ElmPageTest
+    {
+        [Theory]
+        [InlineData("TestFiles/SourceFile.txt")]
+        [InlineData(@"TestFiles\SourceFile.txt")]
```
NTaylorMullen: These are relative paths.

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,210 @@
...
+using System.Reflection;
+using Microsoft.AspNet.FileProviders;
+using Xunit;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    public class ElmPageTest
+    {
+        [Theory]
+        [InlineData("TestFiles/SourceFile.txt")]
```
NTaylorMullen: `MemberData` and re-use with next test.

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,210 @@
...
+                "func1",
+                absoluteFilePath,
+                lineNumber: 10,
+                fileProvider: provider,
+                sourceCodeLineCount: 6);
+
+            // Assert
+            // Lines 4-16 (inclusive) is the code block
+            Assert.Equal(4, stackFrame.PreContextLine);
+            Assert.Equal<string>(GetCodeLines(4, 9), stackFrame.PreContextCode);
```
NTaylorMullen: isn't xunit smart enough to not need `<string>`?

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            // Assert
+            // Lines 1 is the code block
+            Assert.Equal(1, stackFrame.PreContextLine);
+            Assert.Empty(stackFrame.PreContextCode);
+            Assert.Equal("Line1", stackFrame.ContextCode.First());
+            Assert.Empty(stackFrame.PostContextCode);
+        }
+
+        private IEnumerable<string> GetCodeLines(int fromLine, int toLine)
```
NTaylorMullen: `GetExpectedCodeLines`

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,210 @@
...
+                sourceCodeLineCount: 6);
+
+            // Assert
+            // Lines 1 is the code block
+            Assert.Equal(1, stackFrame.PreContextLine);
+            Assert.Empty(stackFrame.PreContextCode);
+            Assert.Equal("Line1", stackFrame.ContextCode.First());
+            Assert.Empty(stackFrame.PostContextCode);
+        }
+
```
NTaylorMullen: Add additional test `lineNumber: 1`, `sourceCodeLineCount: 1`

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: Good point!

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: When the call to `ToArray()` realizes the query.

NTaylorMullen: WHaaaa... Holy cow that's hard to find. It's two levels of indirection. Don't rely on this disposing this backing stream; it's too easy to break in the future.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: In the case when the file provider `exists` fails..for example, one could write a file provider which gets the files from a database or any other source and sets it on the RazorViewEngineOptions but does not set the provider in ErrorPage, in which case the exists check would fail...

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
victorhurdugaci: No, no! Set the array to empty. It is horrible to get a null a back from an IEnumerable. It breaks foreach and you always have to check for it

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
victorhurdugaci: That's the C# compiler, not xunit :)

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: I believe its to handle the case where the stack trace contains some text like `--End of stack trace...` 
https://cloud.githubusercontent.com/assets/6559784/8710577/94596bb8-2aff-11e5-853c-3c4c4ea2afa6.png

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
kichalla: @victorhurdugaci I am not clear about your comment...like how is it being hierarchical here matters...we are already handling the hierarchical paths when we are using `/`?

src/Microsoft.AspNet.Diagnostics/project.json
kichalla: I am using VS RTM version....I think I auto-formatted the document...

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: Sure, good point

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: I wanted it to be static to make the unit testing easier as otherwise I need to setup the the stuff required in the constructor

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
kichalla: This test is for absolute paths where as the other one is for relative paths...it would be clear in case of failure...

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: kk :+1:

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
NTaylorMullen: nah not C# compiler `PreContextCode` is an `IEnumerable<string>` XUnit is the one who has to be smart here.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: We already check for it.... If you want to leave it as an empty array fix the views to understand empty array = don't generate `PreContextCode` HTML.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: Helper methods are your friend (`CreateTestErrorPageMiddleWare(FileProvider fileProvider)`). Making it a static out of convenience isn't a good reason.

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
NTaylorMullen: Hu? It's already semi-unclear that the data to the absolute path method is relative paths. A member will at least abstract that part away.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: Ah interesting

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: Good catch...I think I changed the existing behavior and have changed it in the latest iteration (although I am not sure how to reproduce a multi-line error...do you have any idea?)...I have tests covering these multiline scenarios, so I think we should be good.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: I modified the function signature to accept a IFileInfo instead...the behavior now should be similar to how it would be when we call `File.ReadLines` which internally opens the file stream and disposes it when the iterator internally is disposed which happens with the call to `ToArray`

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: Sure, got it...made the changes...

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
kichalla: I have changed it. (Btw it's C# type inference and not Xunit)

kichalla:
Updated @NTaylorMullen @victorhurdugaci

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,328 @@
...
+using System.Runtime.Versioning;
+using System.Threading.Tasks;
+using Microsoft.AspNet.FileProviders;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.Runtime;
+using Xunit;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    public class ElmPageTest
```
kichalla: This should have been `ErrorPageMiddlewareTest`

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cshtml
```diff
@@ -45,33 +45,27 @@
...
-                                            <li><span>@line</span></li>
-                                        }
-                                    </ol>
-                                }
+                                <ol start="@frame.PreContextLine" class="collapsible">
+									@foreach (var line in frame.PreContextCode)
+									{
+										<li><span>@line</span></li>
+									}
+								</ol>
```
victorhurdugaci: Nit: indentation

kichalla: weird...i used Notepad++(as opening in VS was causing other unrelated changes) for this in which it was ok...will fix..

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: Not familiar with how the `compilationFailure.Messages` are organized so can't say I know an exact reproduction of a multi-line error. I'd imagine C# block code which is spread across multiple lines would be the route to take however.

NTaylorMullen: Could also make this method internal and build your own compilation exceptions to simulate :smile:

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,104 @@ private StackFrame StackFrame(Chunk line)
...
             if (File.Exists(file))
             {
-                var code = File.ReadLines(file);
-                ReadFrameContent(frame, code, lineNumber, lineNumber);
+                lines = File.ReadLines(file);
             }
+            else
+            {
+                // Handle relative paths and embedded files
+                var fileInfo = _options.FileProvider.GetFileInfo(file);
```
NTaylorMullen: `_fileProvider`. Add test to validate `null` `_options.FileProvider` results in a `PhysicalFileProvider` being used.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,104 @@ private StackFrame StackFrame(Chunk line)
...
         }
 
-        private void ReadFrameContent(StackFrame frame,
-                                      IEnumerable<string> code,
-                                      int startLineNumber,
-                                      int endLineNumber)
+        private void ReadFrameContent(
+            StackFrame frame,
+            IEnumerable<string> allLines,
+            int errorStartLineNumberInFile,
```
NTaylorMullen: revert to old names. `startLineNumber`, `endLineNumber`

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,104 @@ private StackFrame StackFrame(Chunk line)
...
+            int errorEndLineNumberInFile)
+        {
+            // Get the line boundaries in the file to be read and read all these lines at once into an array.
+            // This will cause the backing stream to be closed.
```
NTaylorMullen: remove backing stream comment.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,104 @@ private StackFrame StackFrame(Chunk line)
...
+            var preErrorLineNumberInFile = Math.Max(errorStartLineNumberInFile - _options.SourceCodeLineCount, 1);
```
NTaylorMullen: `_options.SourceCodeLineCount` is that the total source code lines? If so, doesn't this change the behavior to always have the entire file as the code block?

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -182,31 +188,104 @@ private StackFrame StackFrame(Chunk line)
...
+            var errorStartLineNumberInArray = errorStartLineNumberInFile - preErrorLineNumberInFile;
+
+            frame.PreContextLine = preErrorLineNumberInFile;
+            frame.PreContextCode = codeBlock.Take(errorStartLineNumberInArray).ToArray();
+            frame.ContextCode = codeBlock
+                .Skip(errorStartLineNumberInArray)
+                .Take(numOfErrorLines)
+                .ToArray();
+            frame.PostContextCode = codeBlock
+                .Skip(errorStartLineNumberInArray)
```
NTaylorMullen: Instead of two just do `.Skip(endLineNumber)`

src/Microsoft.AspNet.Diagnostics/ErrorPageOptions.cs
```diff
@@ -23,5 +25,11 @@ public ErrorPageOptions()
         /// source code referenced by the exception stack trace is present on the server.
         /// </summary>
         public int SourceCodeLineCount { get; set; }
+
+        /// <summary>
+        /// Provides files containing source code which is used to display the line(s) of exception.
```
NTaylorMullen: `Provides files containing source code used to display contextual information of an exception.`

src/Microsoft.AspNet.Diagnostics/ErrorPageOptions.cs
```diff
@@ -23,5 +25,11 @@ public ErrorPageOptions()
...
+        /// if not provided, <see cref="PhysicalFileProvider"/> is used by default.
```
NTaylorMullen: Move this bit into `<remarks>`: `If <c>null</c> <see cref="ErrorPageMiddleware" /> will use <see cref="PhysicalFileProvider"/>.`

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cshtml
```diff
@@ -45,33 +45,27 @@
...
-                                @if (frame.PreContextCode != null)
-                                {
-                                    <ol start="@frame.PreContextLine" class="collapsible">
-                                        @foreach (var line in frame.PreContextCode)
-                                        {
-                                            <li><span>@line</span></li>
-                                        }
-                                    </ol>
-                                }
+                                <ol start="@frame.PreContextLine" class="collapsible">
```
NTaylorMullen: If there's no `PreContextCode` this will still render.

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cshtml
```diff
@@ -45,33 +45,27 @@
...
-                                @if (frame.PostContextCode != null)
-                                {
-                                    <ol start='@(frame.Line + 1)' class="collapsible">
-                                        @foreach (var line in frame.PostContextCode)
-                                        {
-                                            <li><span>@line</span></li>
-                                        }
-                                    </ol>
-                                } 
+                                <ol start='@(frame.Line + 1)' class="collapsible">
```
NTaylorMullen: ditto for `PostContextCode`

src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cshtml
```diff
@@ -94,18 +94,15 @@
...
-                                        @if (frame.PreContextCode != null)
-                                        {
-                                            <ol start="@frame.PreContextLine" class="collapsible">
-                                                @foreach (var line in frame.PreContextCode)
-                                                {
-                                                    <li><span>@line</span></li>
-                                                }
-                                            </ol>
-                                        } 
+                                        <ol start="@frame.PreContextLine" class="collapsible">
```
NTaylorMullen: Same comments as in `CompilationErrorPage.cshtml`

src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cshtml
```diff
@@ -94,18 +94,15 @@
                                     <h3>@frame.Function in <code title="@frame.File">@System.IO.Path.GetFileName(frame.File)</code></h3>
                                 }
 
-                                @if (frame.Line != 0 && frame.ContextCode !=null && frame.ContextCode.Any())
+                                @if (frame.Line != 0 && frame.ContextCode.Any())
```
NTaylorMullen: Q: Why in `CompilationErrorPage.cshtml` do we check `ContextCode != null` and now we don't here?

src/Microsoft.AspNet.Diagnostics/Views/StackFrame.cs
```diff
@@ -32,19 +33,19 @@ public class StackFrame
         public int PreContextLine { get; set; }
 
         /// <summary>
-        /// 
+        /// Lines of code before the actual error line(s)
```
NTaylorMullen: period here and in other added docs.

src/Microsoft.AspNet.Diagnostics/Views/StackFrame.cs
```diff
@@ -32,19 +33,19 @@ public class StackFrame
...
         /// </summary>
-        public IEnumerable<string> PreContextCode { get; set; }
+        public IEnumerable<string> PreContextCode { get; set; } = Enumerable.Empty<string>();
 
         /// <summary>
-        /// 
+        /// Lines having the error
```
NTaylorMullen: `Lines of code responsible for the error.`

src/Microsoft.AspNet.Diagnostics/Views/StackFrame.cs
```diff
@@ -32,19 +33,19 @@ public class StackFrame
         public int PreContextLine { get; set; }
 
         /// <summary>
-        /// 
+        /// Lines of code before the actual error line(s)
         /// </summary>
-        public IEnumerable<string> PreContextCode { get; set; }
+        public IEnumerable<string> PreContextCode { get; set; } = Enumerable.Empty<string>();
```
NTaylorMullen: Why add these? Very odd to declare enumerables as enumerable.empty in poco classes.

test/Microsoft.AspNet.Diagnostics.Tests/ElmPageTest.cs
```diff
@@ -0,0 +1,328 @@
...
+    {
+        public static TheoryData<string> AbsolutePathsData
```
NTaylorMullen: `PathData`, then re-use in both relative and absolute

test/Microsoft.AspNet.Diagnostics.Tests/TestFileProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.FileProviders;
+using Microsoft.Framework.Caching;
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    public class TestFileProvider : IFileProvider
```
NTaylorMullen: Move to `private class` in `ErrorPageMiddlewareTest`, it's not used anywhere else. Ditto with `TestFileInfo.cs`

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: No, `_options.SourceCodeLineCount` is number of lines before, after the lines of error..

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: No, `endLineNumber` or `errorEndLineNumberInFile` (that is the reason I do not want to rename the variable) is the line number  in file...here we are referring to the code block in the array...

src/Microsoft.AspNet.Diagnostics/Views/CompilationErrorPage.cshtml
kichalla: Good point..will fix it...

src/Microsoft.AspNet.Diagnostics/Views/StackFrame.cs
kichalla: This is so that we need not null checks on the views where sometimes contextcode, pre or post can be null

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
NTaylorMullen: Dur, misread this bit of code. This should work though: `.Skip(errorStartLineNumberInArray + numOfErrorLines)`

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: :+1:

src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cshtml
kichalla: Good catch..fixed it

kichalla:
Updated @NTaylorMullen

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -183,31 +189,102 @@ private StackFrame StackFrame(Chunk line)
...
+            }
+            else
+            {
+                return GetStackFrame(function, file, lineNumber, lineNumber);
+            }
         }
 
-        private StackFrame LoadFrame(string function, string file, int lineNumber)
+        // make it internal to enable unit testing
+        internal StackFrame GetStackFrame(string function, string file, int startLineNumber, int endLineNumber)
```
NTaylorMullen: Do we need the `endLineNumber`? Seeing every call to this send the same line number for start and end.

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
```diff
@@ -183,31 +189,102 @@ private StackFrame StackFrame(Chunk line)
...
-                ReadFrameContent(frame, code, lineNumber, lineNumber);
+                lines = File.ReadLines(file);
             }
+            else
+            {
+                // Handle relative paths and embedded files
+                var fileInfo = _fileProvider.GetFileInfo(file);
+                if (fileInfo.Exists)
+                {
+                    // ReadLines doesn't accept a stream. Use ReadLines as its more efficient
```
NTaylorMullen: `... Use File.ReadLines as its more efficient`

src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs
kichalla: right, we need it in the case of compilation errors https://github.com/aspnet/Diagnostics/blob/kiran/razor-sourcecode-lines/src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs#L119

NTaylorMullen: That's `ReadFrameContent` not `GetStackFrame`

NTaylorMullen:
:watch: just one last question to ensure there's not a hidden bug associated with https://github.com/aspnet/Diagnostics/pull/158#discussion_r35591996

NTaylorMullen:
Talked in person, once method name is updated :shipit:

kichalla:
updated @NTaylorMullen

NTaylorMullen:
:shipit:

PR #161, opened 2015-07-30
Minor code improvements

dpalezevic:


src/Microsoft.AspNet.Diagnostics.Entity/DatabaseErrorPageOptions.cs
```diff
@@ -8,7 +8,7 @@ public class DatabaseErrorPageOptions
         public static DatabaseErrorPageOptions ShowAll => new DatabaseErrorPageOptions
                                                               {
                                                                   ShowExceptionDetails = true,
-                                                                  ListMigrations = true,
+                                                                  ListMigrations = true
```
Eilon: We prefer to keep the trailing comma wherever possible because it makes the diffs cleaner when items are added or removed from the list.

dpalezevic: Thanks for the explanation.

Eilon:
@rowanmiller FYI on this because this change affects the EF files. Just code cleanup.

@kichalla can you merge?

rowanmiller:
EF changes look good to me :smile:

kichalla:
commit 4a9029fc9cfefb8ba10dc63b1a233b01e0c5d10a

Thanks @dpalezevic

PR #163, opened 2015-08-04
Add Status Description to UseStatusCodePagesWithReExecute

RehanSaeed:
Currently you can use UseStatusCodePagesWithReExecute to create a path like ```/error/404```. It would be nice if you could add the error status name to the URL like ```/error/404/notfound```.


Tratcher:
I don't think we should do this.:
0) Reason phrases are for human display, not programmatic use. E.g. Feel free to display them in HTML, but don't use them in URLs.
1) Reason phrases are effectively obsolete, HTTP/2 has removed them.
2) The HttpStatusCode does not and cannot define every possible status code/reason.
3) This middleware re-writes the url on the server side, the client never sees it.
4) It makes the downstream implementation more difficult.

Eilon:
Agree with @Tratcher ...

PR #164, opened 2015-08-07
Added Content-Type header to RuntimeInfoPage view and reacting to nam…

kichalla:
…espace change

@victorhurdugaci

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics/Views/RuntimeInfoPage.cshtml
+++ b/src/Microsoft.AspNet.Diagnostics/Views/RuntimeInfoPage.cshtml
@@ -3,7 +3,7 @@
 @using System.Linq
 @using Microsoft.AspNet.Diagnostics
 @using Microsoft.AspNet.Diagnostics.Views
-@using Microsoft.Framework.Runtime;
+@using Microsoft.Dnx.Runtime;
 
 @functions
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Diagnostics/Views/RuntimeInfoPage.cshtml
```diff
@@ -14,6 +14,9 @@
 
     public RuntimeInfoPageModel Model { get; set; }
 }
+@{
+    Response.ContentType = "text/html; charset=utf-8";
+}
 <!DOCTYPE html>
```
Eilon: Can you run the output of the *rendered* page through the [W3C validator](http://validator.w3.org/)? I suspect this page won't validate because the DOCTYPE isn't at the top of the page. (Razor will render a few empty lines before it.)

If the page doesn't validate, please log a new bug with the validation errors and we can triage it separately.

Eilon:
:shipit:

src/Microsoft.AspNet.Diagnostics/Views/RuntimeInfoPage.cshtml
kichalla: @Eilon I now tried validating the generated page with the validator and it didn't indicate any errors (yes, there were 3 empty lines at the beginning of the document)

Eilon: Ah good! :smile:

PR #170, opened 2015-08-17
[Fixes #167] Error page/handler should reset any buffered data, clear …

### LINKED ISSUE #167: Error page/handler should reset any buffered data, clear headers, etc.
RE: https://github.com/aspnet/Hosting/issues/299

If the response is buffering when an exception occurs, the error handler and error page middleware should clear the headers and buffer before taking their recovery actions.

ErrorHandler already clears the headers and sets a status code:
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs#L59-L61

ErrorPage sets the status code but only alters select headers. It doesn't even check Response.HasStarted before trying to do anything, so in this scenario it is already throwing trying to modify the status code and headers.
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cshtml#L16-L19

kichalla:
@Tratcher (fyi...i installed the locally built package of buffering and tested it)

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs
+++ b/src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs
@@ -5,6 +5,7 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
 using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Diagnostics
@@ -60,7 +61,12 @@ namespace Microsoft.AspNet.Diagnostics
                     context.Response.Headers.Clear();
                     context.Response.OnStarting(_clearCacheHeadersDelegate, context.Response);
 
-                    // TODO: Try clearing any buffered data. The buffering feature/middleware has not been designed yet.
+                    // if buffering is enabled, then clear it as data could have been written into it.
+                    if (context.Response.Body.CanSeek)
+                    {
+                        context.Response.Body.SetLength(0);
+                    }
+
                     await _options.ErrorHandler(context);
                     // TODO: Optional re-throw? We'll re-throw the original exception by default if the error handler throws.
                     return;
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs
```diff
@@ -60,7 +61,16 @@ public ErrorHandlerMiddleware(RequestDelegate next, ILoggerFactory loggerFactory
                     context.Response.Headers.Clear();
                     context.Response.OnStarting(_clearCacheHeadersDelegate, context.Response);
 
-                    // TODO: Try clearing any buffered data. The buffering feature/middleware has not been designed yet.
+                    // if buffering is enabled, then clear it as data could have been written into it.
+                    var bufferingFeature = context.GetFeature<IHttpBufferingFeature>();
+                    if (bufferingFeature != null)
+                    {
+                        if (context.Response.Body.CanSeek)
```
kichalla: I wonder if we should even check for the presence of `IHttpBufferingFeature` and just do this check?

Tratcher: No, you don't need to check for the IHttpBufferingFeature, just CanSeek

test/Microsoft.AspNet.Diagnostics.Tests/project.json
```diff
@@ -3,6 +3,7 @@
         "warningsAsErrors": "true"
     },
     "dependencies": {
+        "Microsoft.AspNet.Buffering": "1.0.0",
```
Tratcher: If you check this dependency in we'll have to add the dependency in the CI graph.

Tratcher:
:watch:

kichalla:
updated @Tratcher

src/Microsoft.AspNet.Diagnostics/Views/ErrorPage.cshtml
```diff
@@ -13,10 +13,8 @@
     public ErrorPageModel Model { get; set; }
 }
 @{
-    Response.StatusCode = 500;
     // TODO: Response.ReasonPhrase = "Internal Server Error";
```
Tratcher: Remove this line.

Tratcher:
:shipit:

kichalla:
09b1c0fa6314db316b386c166e07b80cc27ed0c2

PR #177, opened 2015-09-10
Use the new Response.Clear() API.

Tratcher:
@lodejard @davidfowl @kichalla @muratg


davidfowl:
Don't check this in until Travis passes. That way we'll know if aspnetlitedev passes

muratg:
Once we have Travis green, :shipit:

Issue #180, opened 2015-09-14
Publish unhandled errors to telemetry source


Eilon:
Publish this:
1. Unhandled errors - wherever in there's a `catch` that doesn't rethrow (same place that stuff is logged right now, probably)
	- Need to do this in both DeveloperExceptionPage and ErrorHandler

Example in MVC: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Actions/MvcRouteHandler.cs#L73-L78

Eilon:
Related: https://github.com/aspnet/Hosting/issues/354

JunTaoLuo:
Just to confirm, the only *unhandled* exception that needs to be written to telemetry is at [DevelopperExceptionPageMiddleware.cs#L86](https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs#L86) where we just create an exception page. All other instance of `Invoke` in  [DevelopperExceptionPageMiddleware.cs](https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs#L68-95) will end up throwing an exception that bubbles up to hosting.

Similarly [ExceptionHandlerMiddleware.cs](https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L35-79)  will throw unhandled exceptions up to hosting or handle it at [ExceptionHandlerMiddleware.cs#L65](https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L65)

cc @Tratcher @Eilon

Tratcher:
Correct, you only need telemetry for app exceptions that are not re-thrown.

PR #185, opened 2015-10-02
Adding telemetry publish for unhandled exceptions to developer exception page

JunTaoLuo:
For #180 cc @Eilon @Tratcher

### DIFF
--- a/test/Microsoft.AspNet.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
+++ b/test/Microsoft.AspNet.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
@@ -3,16 +3,20 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.Tracing;
 using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.Versioning;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Diagnostics.Views;
 using Microsoft.AspNet.FileProviders;
+using Microsoft.AspNet.TestHost;
 using Microsoft.AspNet.Testing;
 using Microsoft.Dnx.Runtime;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Xunit;
@@ -298,7 +302,8 @@ namespace Microsoft.AspNet.Diagnostics
                 (httpContext) => { return Task.FromResult(0); },
                 errorPageOptions,
                 new LoggerFactory(),
-                new TestApplicationEnvironment());
+                new TestApplicationEnvironment(),
+                new TelemetryListener("Microsoft.Aspnet"));
 
             return middleware;
         }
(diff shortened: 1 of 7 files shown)


test/Microsoft.AspNet.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
```diff
@@ -471,5 +477,72 @@ public class ErrorData
...
+            telemetryListener.SubscribeWithAdapter(listener);
+
+            // Act
+            await server.CreateClient().GetAsync("/path");
+            
+            // Assert
+            Assert.NotNull(listener.DiagnosticUnhandledException?.HttpContext);
+            Assert.NotNull(listener.DiagnosticUnhandledException?.Exception);
+            Assert.Null(listener.HostingUnhandledException?.HttpContext);
+            Assert.Null(listener.HostingUnhandledException?.Exception);
```
Tratcher: Assert NotNull HostingEndRequest

JunTaoLuo: I'm guessing I should Assert NotNull HostingBeginRequest as well?

Tratcher:
@avanderhoorn @rynowak

Tratcher:
Include the bug number in the commit message.

test/Microsoft.AspNet.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
Tratcher: *shrug* that's out of scope of this change.

JunTaoLuo:
@Tratcher I will add the bug number once I squash all commits

avanderhoorn:
When is this event triggered as apposed to `"Microsoft.AspNet.Hosting.UnhandledException"` [here](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs#L119).

JunTaoLuo:
@avanderhoorn When the developer exception page successfully displays. Since the exception will be suppressed, we want to publish it to telemetry source as diagnostic unhandled exception. If for other reasons displaying the exception fails, the exception will be bubbled up to hosting and will trigger the telemetry publish there.

avanderhoorn:
Great, so in the case that this event happens, we wont get the `hosting` one.

JunTaoLuo:
@avanderhoorn right.

avanderhoorn:
Makes sense. Thanks! :shipit:

Tratcher:
We also need this here: https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L66

JunTaoLuo:
@Tratcher I was asking for that in my previous comment in the issue #180 . That error is technically handled and I wasn't sure if that should also publish to telemetry.

avanderhoorn:
For our part we would like to know about that exception too.

JunTaoLuo:
@avanderhoorn cool. I added the telemetry publish for exception handler as well. It's going to publish to `Microsoft.AspNet.Diagnostics.HandledException` just to differentiate from the unhandled exception from developer exception page, is that good?

avanderhoorn:
Sounds great!

JunTaoLuo:
I'm going to take that as a :shipit: All tests passed so I'll merge to dev.

avanderhoorn:
Thanks!

PR #188, opened 2015-10-09
Replace NotNullAttribute with thrown exceptions

pranavkm:


src/Microsoft.AspNet.Diagnostics.Elm/ElmScope.cs
```diff
@@ -5,7 +5,7 @@
 #else
 using System.Threading;
 #endif
-using Microsoft.Extensions.Internal;
+
```
Eilon: This

Eilon:
:shipit:

PR #189, opened 2015-10-13
Rybrande/error page copyable

ryanbrandenburg:
The core ask here is to add an area on the error page where it's easy to copy the stack-trace.

However, in order to do that in a sane way we needed to fix up the PageGenerator utility to work with new versions of Razor, and while we were at it we made it build the pages in the makefile so we never get back to a state of quietly failing to generate those.

The View/*.cs changes are all generated by the above process.


src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
```diff
@@ -14,56 +14,89 @@
...
-    var childIndex = 0;
-    while (messageIndex < node.Messages.Count && childIndex < node.Children.Count)
-    {
-        if (node.Messages[messageIndex].Time < node.Children[childIndex].StartTime)
-        {
-            @LogRow(node.Messages[messageIndex])
-            messageIndex++;
-        }
-        else
+        return new HelperResult((__razor_helper_writer) =>
```
Eilon: Weeiiiird, what's a `__razor_helper_writer`? Can this have a simpler name?

src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
```diff
@@ -14,56 +14,89 @@
...
+            if (log.Severity >= Model.Options.MinLevel &&
+                (string.IsNullOrEmpty(Model.Options.NamePrefix) || log.Name.StartsWith(Model.Options.NamePrefix, StringComparison.Ordinal)))
+            {
+
+                WriteLiteralTo(__razor_helper_writer, "        <tr>\r\n            <td>");
+                WriteTo(__razor_helper_writer, string.Format("{0:MM/dd/yy}", log.Time));
+                WriteLiteralTo(__razor_helper_writer, "</td>\r\n            <td>");
+                WriteTo(__razor_helper_writer, string.Format("{0:H:mm:ss}", log.Time));
+                WriteLiteralTo(__razor_helper_writer, "</td>\r\n            <td");
+                WriteAttributeTo(__razor_helper_writer, "class", Tuple.Create(" class=\"", 768), Tuple.Create("\"", 819),
```
Eilon: Is all this stuff just to work around the lack of `@helper`? All this `Tuple()` stuff is probably not needed, is it?

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
```diff
@@ -117,6 +118,48 @@ protected void WriteLiteral(object value)
                 values);
         }
 
+        private List<string> _attributeValues { get; set; }
```
Eilon: `AttributeValues`?

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
```diff
@@ -117,6 +118,48 @@ protected void WriteLiteral(object value)
...
+
+        protected void WriteAttributeValue(string thingy, int startPostion, object value, int endValue, int dealyo, bool yesno)
+        {
+            if(_attributeValues == null)
```
Eilon: Format document <kbd>Ctrl</kbd>+<kbd>K</kbd> <kbd>Ctrl</kbd>+<kbd>D</kbd>.

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
```diff
@@ -117,6 +118,48 @@ protected void WriteLiteral(object value)
...
+            {
+                _attributeValues = new List<string>();
+            }
+
+            _attributeValues.Add(value.ToString());
+        }
+
+        private string _attributeEnding { get; set; }
```
Eilon: `AttributeEnding`

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
```diff
@@ -117,6 +118,48 @@ protected void WriteLiteral(object value)
...
+
+            Output.Write(begining);
+            _attributeEnding = ending;
+        }
+
+        protected void EndWriteAttribute()
+        {
+            if(string.IsNullOrEmpty(_attributeEnding))
+            {
+                throw new InvalidOperationException();
```
Eilon: Is this basically an assertion?

src/PageGenerator/project.json
```diff
@@ -5,18 +5,11 @@
...
   },
   "dependencies": {
-    "Microsoft.AspNet.Diagnostics.Elm": "1.0.0-*",
-    "Microsoft.AspNet.Diagnostics.Entity": "7.0.0-*",
     "Microsoft.AspNet.Razor": "4.0.0-*"
   },
 
   "frameworks": {
-    "dnx451": { },
-    "dnxcore50": {
```
Eilon: Can this not run on CoreCLR anymore? It should be able to...

src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
ryanbrandenburg: Basically yeah, this is my attempt at a direct switch between the two without changing much. I wanted to minimize my changes, but if we want I can go in and clean those up using some of the Write and WriteLiteral overloads.

src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
ryanbrandenburg: Actually, now that I reflect on it we can probably get away without that HelperResult (and thus __razor_helper_writer) at all now that we have the Write and WriteLiteral overloads, which figure out what to write to for you. I'll give that a shot, and if it doesn't work I'll make up a better name.

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
ryanbrandenburg: I was under the impression that private members were [_camelCase](https://github.com/aspnet/Home/wiki/Engineering-guidelines#coding-style-guidelines--general)? Or is that only on fields and not properties?

src/PageGenerator/project.json
ryanbrandenburg: We got a problem with assembly versioning on Razor's version of System.Runtime when we allowed CoreCLR. @pranavkm suggested we just remove it since this is a tool rather than a library.

src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
Eilon: Yeah at a minimum a better name, but if it can just be cleaned up in general, that would be best.

src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
Eilon: Sounds good. I think at the very list get rid of the Tuple stuff, but if you can make it squeaky clean that would be great.

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
Eilon: All properties are PascalCase. All instance fields are always _underscoreCamelCase.

Eilon: I'll fix the guidelines, not sure why it said that.

src/PageGenerator/project.json
Eilon: It's a tool, but it needs to be able to run cross-plat, especially now that it runs as part of the build, and some platforms have only CoreCLR.

@pranavkm do you have more info on what was broken?

pranavkm: I guess we could run it only on CoreCLR instead. There was an issue with a version mismatch. Might just need to be hoisted here.

src/Microsoft.AspNet.Diagnostics/WelcomePage/Views/WelcomePage.cs
```diff
@@ -6,6 +6,12 @@ namespace Microsoft.AspNet.Diagnostics.Views
 #line default
 #line hidden
     ;
+#line 2 "WelcomePage.cshtml"
```
pranavkm: Could we not check in these files and rely on them being generated?

src/PageGenerator/project.json
Eilon: It seems odd to me that we can't write an app that uses Razor that runs on both CLRs...

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.cshtml
```diff
@@ -127,6 +128,16 @@
                         }
                         </ul>
                     </li>
+                    <li>
+                        <div class="rawExceptionBlock">
+                            <div>
+                                <a role="button" id="rawExceptionButton">Show Copyable Exception</a>
```
Tratcher: We should use similar phrasing as the startup exception page: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Startup/StartupExceptionPage.cs#L64

Tratcher: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Startup/StartupExceptionPage.cs#L73

Tratcher:
Screenshot?

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
ryanbrandenburg: I'm guessing you mean to replace that block with Debug.Assert(string.IsNullOrEmpty(AttributeEnding), "AttributeEnding can't be null or empty") ?

src/Microsoft.AspNet.Diagnostics/WelcomePage/Views/WelcomePage.cs
ryanbrandenburg: I would definitely support that change if it weren't for the fact that they aren't built when you just run from VS, only when you build.cmd. Not including them would therefore confuse/mess up people who haven't run build.cmd.

src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
Eilon: It's odd to throw an exception with no message. So what I'm asking is whether this is effectively an assertion (as in, "this can't possibly happen"), or if there is a reasonable way where this *could* happen, in which case it should have a message.

If it's really more of an assertion then this should just be an assert, yes.

ryanbrandenburg:
developerexceptionscreenshot
@Tratcher here's a screenshot.

makefile.shade
```diff
@@ -5,4 +5,8 @@ var FULL_VERSION='0.1'
 var AUTHORS='Microsoft Open Technologies, Inc.'
 
 use-standard-lifecycle
-k-standard-goals
\ No newline at end of file
+k-standard-goals
+
+#generatepages target='initialize'
+	k command='run ../Microsoft.AspNet.Diagnostics' dnxDir='src/PageGenerator'
+
```
pranavkm: Is this whitespace?

src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
```diff
@@ -14,56 +15,86 @@
...
-        }
+            if (log.Severity >= Model.Options.MinLevel &&
+               (string.IsNullOrEmpty(Model.Options.NamePrefix) || log.Name.StartsWith(Model.Options.NamePrefix, StringComparison.Ordinal)))
+            {
+                WriteLiteralTo(writer, "        <tr>\r\n            <td>");
+                WriteTo(writer, string.Format("{0:MM/dd/yy}", log.Time));
+                WriteLiteralTo(writer, "</td>\r\n            <td>");
+                WriteTo(writer, string.Format("{0:H:mm:ss}", log.Time));
+                WriteLiteralTo(writer, "</td>\r\n            <td");
+                WriteAttributeTo(writer, "class", " class=\"", "\"",
```
pranavkm: This is fine, but you could simply replace these with `WriterLiteralTo` \ `Write` method and remove all the `\r\n` while you're at it. It'd make it marginally easier to read:

```C#

WriteLiteralTo(writer, "<tr>");
WriteLiteralTo(writer, "<td>");
...
WriteLiteralTo(writer, "</td><td class=\"");
WriteTo(writer, log.Severity.ToString().ToLowerInvariant());
WriteLiteralTo("\" </td>");
```

ryanbrandenburg: The problem with removing all the whitespace there is it makes the resulting file super hard to read. As for replacing the WriteAttributeTo calls, I'm ok with that unless someone else has a problem with it.

Eilon: I'd kind of like to keep the whitespace. Makes the code here ugly, but better for society :smile:

src/Microsoft.AspNet.Diagnostics.Elm/project.json
```diff
@@ -17,5 +16,6 @@
                 "System.Threading": "4.0.11-beta-*"
             }
         }
-    }
+    },
+    "resource": ["Views/*.cshtml"]
```
Eilon: Resources? Naah, that means they'll get embedded as files in the DLL, which we definitely don't want. I think you just want `compileExclude`: https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/project.json#L116-L122

src/Microsoft.AspNet.Diagnostics/project.json
```diff
@@ -27,5 +27,6 @@
                 "System.Reflection.Extensions": "4.0.1-beta-*"
             }
         }
-    }
+    },
+    "resource": ["DeveloperExceptionPage/Views/*.cshtml", "RuntimeInfo/Views/*.cshtml", "WelcomePage/Views/*.cshtml"]
```
Eilon: See previous comment about `compileExclude`.

src/Microsoft.AspNet.Diagnostics.Elm/project.json
Eilon: Or actually, what "bad" thing happens if it's left alone? Why mention it at all?

ryanbrandenburg: If left in it seems the build process tries to treat them as real cshtml files, and since they're not this results in errors like the one [https://travis-ci.org/aspnet/Diagnostics/builds/85435845](here). I'm good moving to compileExclude, but might it be even better to just change the extension to not be cshtml since they're really not?

src/Microsoft.AspNet.Diagnostics.Elm/Views/DetailsPage.cshtml
pranavkm: :crying_cat_face:

src/Microsoft.AspNet.Diagnostics.Elm/project.json
Eilon: Ah weird. Well, `compileExclude` is I think the right thing to do here. I wouldn't rename the file extension because they *are* in fact real Razor files (they use Razor syntax, after all). The fact that they're used for non-MVC scenarios is separate.

ryanbrandenburg:
All better?

Eilon:
I'm gonna go with :shipit: !

PR #192, opened 2015-10-19
:bug: No database error page for existing databases

### LINKED ISSUE #182: DatabaseErrorPage: Review what is displayed when migrations are not used
per https://github.com/aspnet/EntityFramework/issues/3152#issuecomment-141568079 there may be some cases where we are showing migrations info when the context targets and existing database and is not using migrations.

rowanmiller:
When there is an existing database and no migrations are present, then we should just display the standard error page as they are not using migrations and need to adjust their mapping code to match the existing schema.
Resolve #182


rowanmiller:
@bricelam could you review this one

bricelam:
:shipit:

PR #197, opened 2015-10-28
Fixing build: handle null messages

JunTaoLuo:
@Tratcher


test/Microsoft.AspNet.Diagnostics.Entity.FunctionalTests/DatabaseErrorPageMiddlewareTest.cs
```diff
@@ -284,6 +284,7 @@ public ApplyMigrationsMiddleware(RequestDelegate next)
         [Fact]
         public async Task Pass_thru_when_context_not_in_services()
         {
+            System.Diagnostics.Debugger.Launch();
```
davidfowl: Wait what?

JunTaoLuo:
Actually we shouldn't even be running into this issue. Need to find out why messages are null.

PR #202, opened 2015-11-04
Strong name everything except Microsoft.AspNet.Diagnostics.Entity.

cesarbs:
https://github.com/aspnet/dnx/issues/2615

cc @davidfowl @Eilon @muratg @moozzyk @pranavkm @Tratcher


cesarbs:
Not signing Microsoft.AspNet.Diagnostics.Entity at this point because EF is not signed yet.

pranavkm:
:shipit:

PR #204, opened 2015-11-09
 * Fix scrolling on error page

ryanbrandenburg:
Only "real" change here is in ErrorPage.js, the rest is just generated.


src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.cs
```diff
@@ -57,21 +57,21 @@ public CompilationErrorPage()
 #line default
 #line hidden
 
-            WriteLiteral("<!DOCTYPE html>\r\n<html>\r\n    <head>\r\n        <meta charset=\"utf-8\" />\r\n        <t" +
-"itle>");
+            WriteLiteral("<!DOCTYPE html>\n<html>\n    <head>\n        <meta charset=\"utf-8\" />\n        <title" +
```
NTaylorMullen: Well something happened here. Did we forget to re-generate the code file or did you generate this on a Linuxish box? :smile:

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.js
```diff
@@ -169,10 +169,11 @@
     $(".page").hide();
     $("#stackpage").show();
 
-    $("#rawExceptionButton").click(function () {
+    $("#rawExceptionButton").click(function (event) {
         var div = document.getElementById('rawException');
         div.style.display = 'inline-block';
         div.scrollIntoView(true);
+        event.preventDefault();
```
NTaylorMullen: Any reason to not also return false? It's been a while since I've JavaScripted :trollface:  but that's the recommended old pattern I'm used to (to do both).

NTaylorMullen:
:watch:

ryanbrandenburg:
:up::date:

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.js
```diff
@@ -169,10 +169,13 @@
     $(".page").hide();
     $("#stackpage").show();
 
-    $("#rawExceptionButton").click(function () {
+    $("#rawExceptionButton").click(function (event) {
```
NTaylorMullen: Did a little bit more research on this. You only need to `return false`. Remove the `event` and its `event.preventDefault()`. 

jQuery handles `return false` as do both `event.preventDefault()` AND `event.stopPropagation()`

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.js
Eilon: @pranavkm FYI because he was the one saying `preventDefault`.

pranavkm: If this was vanilla JS `event.preventDefault()` would have been the right thing to do. It's ok to return false in a jquery handled event: http://www.mail-archive.com/jquery-en@googlegroups.com/msg71371.html

Eilon: Ok so we really should clean this up and only `return false`.

pranavkm: Ok, looking at the rest of the codes, we're not using jquery. We're would be the right thing.

Eilon: Oh duh, yeah, of course it doesn't use jQuery :smile: Ok...

NTaylorMullen: O_o. The fact that it's not jQuery is suuuuuuupppPpPPpeEerrRrr lame given how we're coding this. I officially hate `NodeCollection` but do like how it isn't jQuery (too heavy). ALrigh alright, `event.preventDefault()` and `event.stopPropagation()`.

Filed an issue to address making this code non-lamish: https://github.com/aspnet/Diagnostics/issues/205

Eilon: Yeah we should look at cleaning that up.

ryanbrandenburg:
:up: :date:

NTaylorMullen:
:shipit:

Eilon:
:shipit:

PR #207, opened 2015-11-18
Add middleware analysis tools, sample.

Tratcher:
Hosting#318 @avanderhoorn @davidfowl @lodejard 
Depends on:
https://github.com/aspnet/Hosting/commit/f8d2b9bcfb0ea3b77631c72526c13b4e774ec7c8
https://github.com/aspnet/HttpAbstractions/commit/0e03acfc5e54d81e7a2616882e8ef97299ed404a

Sample output:
```
info: Microsoft.AspNet.Hosting.Internal.HostingEngine[1]
      Request starting HTTP/1.1 GET http://localhost:5000/map/foo
MiddlewareStarting: RequestServicesContainerMiddleware; /map/foo
MiddlewareStarting: DeveloperExceptionPageMiddleware; /map/foo
MiddlewareStarting: IISPlatformHandlerMiddleware; /map/foo
MiddlewareStarting: UseInlineMiddleware; /map/foo
MiddlewareStarting: MapMiddleware; /map/foo
MiddlewareStarting: RunInlineMiddleware; /foo
MiddlewareFinished: RunInlineMiddleware; 200
MiddlewareFinished: MapMiddleware; 200
MiddlewareFinished: UseInlineMiddleware; 200
MiddlewareFinished: IISPlatformHandlerMiddleware; 200
MiddlewareFinished: DeveloperExceptionPageMiddleware; 200
MiddlewareFinished: RequestServicesContainerMiddleware; 200
info: Microsoft.AspNet.Hosting.Internal.HostingEngine[2]
      Request finished in 0.0047ms 200
verb: Microsoft.Net.Http.Server.WebListener[0]
      Sending Response: HTTP/1.1 200 OK; Headers: { Content-Length: 11; }
```

### DIFF
--- a/samples/MiddlewareAnalysisSample/Startup.cs
+++ b/samples/MiddlewareAnalysisSample/Startup.cs
@@ -1,0 +1,106 @@
+using System;
+using System.Diagnostics;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DiagnosticAdapter;
+using Microsoft.Extensions.Logging;
+
+namespace MiddlewareAnaysisSample
+{
+    public class Startup
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddMiddlewareAnalysis();
+        }
+
+        public void Configure(IApplicationBuilder app, ILoggerFactory factory, DiagnosticListener diagnosticListener)
+        {
+            // Displays all log levels
+            factory.AddConsole(LogLevel.Verbose);
+
+            // Listen for middleware events and log them to the console.
+            var listener = new TestDiagnosticListener();
+            diagnosticListener.SubscribeWithAdapter(listener);
+
+            // Build our application pipeline
+
+            // Named via app.UseMiddleware<T>()
+            app.UseDeveloperExceptionPage();
+            app.UseIISPlatformHandler();
+
+            // Anonymous method inline middleware
+            app.Use((context, next) =>
+            {
+                // No-op
+                return next();
+            });
+
+            app.Map("/map", subApp =>
+            {
+                subApp.Run(context =>
+                {
+                    return context.Response.WriteAsync("Hello World");
+                });
+            });
+
+            // Low level anonymous method inline middleware, named Diagnostics.Middleware.Analysis.Startup+<>c by default
+            app.Use(next =>
+            {
+                return context =>
+                {
+                    return next(context);
+                };
+            });
+            
+            app.Map("/throw", throwApp =>
+            {
+                throwApp.Run(context => { throw new Exception("Application Exception"); });
+            });
+
+            // The home page.
+            app.Properties["analysis.NextMiddlewareName"] = "HomePage";
+            app.Use(async (context, next) =>
+            {
+                if (context.Request.Path == "/")
+                {
+                    context.Response.ContentType = "text/html";
+                    await context.Response.WriteAsync("<html><body>Welcome to the sample<br><br>\r\n");
+                    await context.Response.WriteAsync("Click here to take a side branch: <a href=\"/map/foo\">Map</a><br>\r\n");
+                    await context.Response.WriteAsync("Click here to throw an exception: <a href=\"/throw\">Throw</a><br>\r\n");
+                    await context.Response.WriteAsync("Click here to for a 404: <a href=\"/404\">404</a><br>\r\n");
+                    await context.Response.WriteAsync("</body></html>\r\n");
+                    return;
+                }
+                else
+                {
+                    await next();
+                }
+            });
+
+            // Note there's always a default 404 middleware at the end of the pipeline.
+        }
+
+        public class TestDiagnosticListener
+        {
+            [DiagnosticName("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareStarting")]
+            public virtual void OnMiddlewareStarting(HttpContext httpContext, string name)
+            {
+                Console.WriteLine($"MiddlewareStarting: {name}; {httpContext.Request.Path}");
+            }
+
+            [DiagnosticName("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareException")]
+            public virtual void OnMiddlewareException(Exception exception, string name)
+            {
+                Console.WriteLine($"MiddlewareException: {name}; {exception.Message}");
+            }
+
+            [DiagnosticName("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareFinished")]
+            public virtual void OnMiddlewareFinished(HttpContext httpContext, string name)
+            {
+                Console.WriteLine($"MiddlewareFinished: {name}; {httpContext.Response.StatusCode}");
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisBuilder.cs
+++ b/src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisBuilder.cs
@@ -1,0 +1,65 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.MiddlewareAnalysis
+{
+    public class AnalysisBuilder : IApplicationBuilder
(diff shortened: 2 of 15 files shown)


Tratcher:
*ping* @davidfowl @avanderhoorn 
/cc: @muratg

davidfowl:
NameNextMiddleware is bizzare doesn't feel right. Don't like it

davidfowl:
So everything else is good except for `AnalysisApplicationBuilderExtensions.GetNextMiddlewareName(this)` and the whole concept of naming anonymous middleware. Just use the name of the delegate (like reflector).

Tratcher:
NameNextMiddleware isn't just for anonymous middleware, It also lets you assign more meaningful names to any other middleware in your app. E.g. like when Identity has three different cookie auth middleware.

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisMiddleware.cs
```diff
@@ -0,0 +1,50 @@
...
+            _next = next;
+            _diagnostics = diagnosticSource;
+            _middlewareName = middlewareName;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            if (_diagnostics.IsEnabled("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareStarting"))
+            {
+                _diagnostics.Write("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareStarting", new { name = _middlewareName, httpContext = httpContext, tickCount = Environment.TickCount });
```
lodejard: Don't put TickCount as event data - assume that the listeners will be using their own clocking system

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisMiddleware.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+
+            try
```
lodejard: change to 
`
//write("starting")
try
{
  await _next(httpContext);
}
catch(exception)
{
  //write("exception")
  throw;
}
finally
{
  //write("finished")
}
`

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisMiddleware.cs
```diff
@@ -0,0 +1,50 @@
...
+                if (_diagnostics.IsEnabled("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareFinished"))
+                {
+                    _diagnostics.Write("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareFinished", new { name = _middlewareName, httpContext = httpContext, tickCount = Environment.TickCount });
+                }
+            }
+            catch (Exception ex)
+            {
+                if (_diagnostics.IsEnabled("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareException"))
+                {
+                    _diagnostics.Write("Microsoft.AspNet.MiddlewareAnalysis.MiddlewareException", new { name = _middlewareName, exception = ex, tickCount = Environment.TickCount });
```
lodejard: shouldn't the exception get the httpcontext also?

samples/MiddlewareAnalysisSample/Startup.cs
```diff
@@ -0,0 +1,106 @@
...
+            services.AddMiddlewareAnalysis();
+        }
+
+        public void Configure(IApplicationBuilder app, ILoggerFactory factory)
+        {
+            // Displays all log levels
+            factory.AddConsole(LogLevel.Verbose);
+
+            // Listen for middleware events and log them to the console.
+            var diagnosticListener = app.ApplicationServices.GetRequiredService<DiagnosticListener>();
```
lodejard: Take `DiagnosticListener` as an additional Configure parameter?

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            return new AnalysisBuilder(InnerBuilder.New());
+        }
+
+        public IApplicationBuilder Use(Func<RequestDelegate, RequestDelegate> middleware)
+        {
+            string middlewareName = AnalysisApplicationBuilderExtensions.GetNextMiddlewareName(this)
+                ?? middleware.Target.ToString(); // Class.Method
+
+            InnerBuilder.UseMiddleware<AnalysisMiddleware>(middlewareName);
```
lodejard: Nice use of middleware constructor parameters!

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class AnalysisServiceCollectionExtensions
+    {
+        public static IServiceCollection AddMiddlewareAnalysis(this IServiceCollection services)
+        {
+            // This should prevent AnalysisStartupFilter from being registered more than once.
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IStartupFilter, AnalysisStartupFilter>());
```
lodejard: Minor suggestion - might be better as transient - othewise the `AnalysisStartupFilter` instance will be added to the application scope service dictionary for the lifetime of the web app

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisMiddleware.cs
Tratcher: That's not how we've been doing it anywhere else. E.g. https://github.com/aspnet/Hosting/pull/463/files#diff-aa52cd5a6af48a03a569c8171c3e02abR74

Tratcher:
Updated.

Tratcher:
I changed the default name from `Func<RequestDelegate, RequestDelegate>.Target` to the resulting `RequestDelegate.Target`. This bypassed most of the indirection issues. The only common broken scenario is the `Use(Func<HttpContext, Func<Task>, Task>)` extension. 
https://github.com/aspnet/HttpAbstractions/commit/dc9529c869a44cdf147bfbfccd7a8c19165bd035#diff-54d6cacc4eeb812168e08000576e7b51R21.  I kept the naming convention to deal with issues like this, but removed all of the sugar for it.

Tratcher:
*poke* @avanderhoorn @davidfowl @muratg

davidfowl:
Why do we need to keep the naming convention?

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisBuilder.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public IFeatureCollection ServerFeatures
+        {
+            get { return InnerBuilder.ServerFeatures;}
+        }
+
+        public RequestDelegate Build()
+        {
+            // Add one maker at the end before the default 404 middleware (or any fancy Join middleware).
+            return InnerBuilder.UseMiddleware<AnalysisMiddleware>("EndOfPipeline")
```
davidfowl: If you call Build twice won't it add the AnalysisMiddleware again?

Tratcher: True. We don't ever call build twice though, and worst case you end up with some duplicate events.

Tratcher: This Build is also only ever called on branches that have first called New (e.g. Map). See https://github.com/aspnet/Diagnostics/pull/207/files#diff-9ac90184c9cfc4a15d6622dbce5b5cf6R19

davidfowl: We don't but the API is busted here.

Tratcher: Would you prefer an InvalidOperationException if it's called twice? That seems worse than a duplicate EndOfPipeline event.

I don't think we can do anything else but change from wrapping the inner builder to completely replacing it.

davidfowl: Ok.

davidfowl:
:shipit:

src/Microsoft.AspNet.MiddlewareAnalysis/AnalysisBuilder.cs
davidfowl: File a bug

davidfowl:
Why is the CI failing?

PR #209, opened 2015-11-18
Error page js cleaning and clarifying

ryanbrandenburg:
As usual in this project, the *ErrorPage.cs files are generated.

Unfortunately all the functions here were used other than toggleClass.


samples/.gitignore
```diff
@@ -0,0 +1 @@
+*/Properties
```
NTaylorMullen: ?

NTaylorMullen:
Are these used: https://github.com/aspnet/Diagnostics/blob/205f5a6ffb96f85d5c91574a6811d3780db4b404/src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.js#L19-L23

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.js
```diff
@@ -1,7 +1,7 @@
 ﻿(function (window, undefined) {
     "use strict";
 
-    function $(selector, element) {
+    function NS(selector, element) {
```
NTaylorMullen: `ns` (javascript convention)

pranavkm:
I was kinda hoping this would remove all the fluent ness from the code. Plus we should be fine dropping IE8 (maybe IE9) support too. :watch:

pranavkm:
Something along these lines: https://gist.github.com/pranavkm/f78e50e7c61ef7d3f8d2

samples/.gitignore
ryanbrandenburg: Running the samples caused them to create property folders, which showed up in git status. That kind of thing drives me bonkers, so I hid them.

ryanbrandenburg:
As per in person conversations with Eilon we're sticking with the current fluent-ness and just trimming it down.

ryanbrandenburg:
:up: :date:

samples/.gitignore
NTaylorMullen: Odd, do you see the same thing when you run the MVC Samples? We don't have this restriction there.

Eilon: Could probably revert this for now. We can always add it back later if it seems more definitely needed.

Eilon:
I think just revert the ignore file, then :shipit:

NTaylorMullen:
:shipit: once you've addressed @Eilon's comments.

PR #215, opened 2015-12-01
Fixing tests to work in Mono

pranavkm:


kichalla:
:shipit:

BrennanConroy:
Still fails on Travis

kichalla:
I believe this is because earlier these tests were not enabled for CoreClr but looks like @pranavkm's recent change added it in project.json

pranavkm:
Yup. I was trying to use Travis to test the PR. Trying it out locally now.

pranavkm:
Updated and enabled tests to run on CoreCLR.

PR #223, opened 2015-12-07
Move ICompilationException to Diagnostics

### LINKED ISSUE #220: Move ICompilationException to Diagnostics.Abstractions
We use `ICompilationException` as a contract between diagnostics and other middlewares (primarily Mvc) that perform runtime compilation to show a pretty printed compilation failure. Currently the contract for this exception is located in `Microsoft.Extensions.CompilationAbstractions`. This would allow removing the CompilationAbstractions dependency.

pranavkm:
Fixes #220

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics.Abstractions/DiagnosticMessage.cs
+++ b/src/Microsoft.AspNet.Diagnostics.Abstractions/DiagnosticMessage.cs
@@ -1,0 +1,64 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    /// <summary>
+    /// A single diagnostic message.
+    /// </summary>
+    public class DiagnosticMessage
+    {
+        public DiagnosticMessage(
+            string message,
+            string formattedMessage,
+            string filePath,
+            int startLine,
+            int startColumn,
+            int endLine,
+            int endColumn)
+        {
+            Message = message;
+            SourceFilePath = filePath;
+            StartLine = startLine;
+            EndLine = endLine;
+            StartColumn = startColumn;
+            EndColumn = endColumn;
+            FormattedMessage = formattedMessage;
+        }
+
+        /// <summary>
+        /// Path of the file that produced the message.
+        /// </summary>
+        public string SourceFilePath { get; }
+
+        /// <summary>
+        /// Gets the error message.
+        /// </summary>
+        public string Message { get; }
+
+        /// <summary>
+        /// Gets the one-based line index for the start of the compilation error.
+        /// </summary>
+        public int StartLine { get; }
+
+        /// <summary>
+        /// Gets the zero-based column index for the start of the compilation error.
+        /// </summary>
+        public int StartColumn { get; }
+
+        /// <summary>
+        /// Gets the one-based line index for the end of the compilation error.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Diagnostics.Abstractions/DiagnosticMessage.cs
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Diagnostics
+{
+    /// <summary>
+    /// A single diagnostic message.
+    /// </summary>
+    public class DiagnosticMessage
+    {
+        public DiagnosticMessage(
```
pranavkm: I removed some of the properties from this - `DiagnosticMessageSeverity`, `ErrorCode` and `object source`. Not sure if there's value in keeping the first - the latter two were used in tooling scenarios.

pranavkm:
cc @davidfowl

src/Microsoft.AspNet.Diagnostics/project.json
```diff
@@ -14,9 +14,9 @@
     "Microsoft.AspNet.FileProviders.Physical": "1.0.0-*",
     "Microsoft.AspNet.Http.Extensions": "1.0.0-*",
     "Microsoft.AspNet.WebUtilities": "1.0.0-*",
-    "Microsoft.Extensions.CompilationAbstractions": "1.0.0-*",
     "Microsoft.Extensions.Logging.Abstractions": "1.0.0-*",
     "Microsoft.Extensions.OptionsModel": "1.0.0-*",
+    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-*",
```
davidfowl: Don't use any types from here. Copy those over too.

pranavkm: There're types in here which use `ILibraryManager` - https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/RuntimeInfo/RuntimeInfoMiddleware.cs#L81. Suggestions on what to do with that?

davidfowl: Ah, that's ok for now.

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -4,19 +4,16 @@
...
 using System.Linq;
 using System.Threading.Tasks;
-using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Diagnostics.Views;
 using Microsoft.AspNet.FileProviders;
 using Microsoft.AspNet.Http;
-using Microsoft.Extensions.CompilationAbstractions;
-using Microsoft.Extensions.PlatformAbstractions;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.PlatformAbstractions;
```
davidfowl: What is used from this namespace?

pranavkm: `IAppEnv` for the base path. We can replace it with code

davidfowl: That;s ok

davidfowl:
:shipit:

PR #229, opened 2015-12-11
 * Fix ErrorPage tabs.

### LINKED ISSUE #228: Developer exception page tabs broken
When you switch tabs on the developer exception page you get a blank screen then even the original tab doesn't show.

ryanbrandenburg:
Fixes #228

As always those first two files are generated.


ryanbrandenburg:
@Tratcher this is from what you showed me.

Tratcher:
@davidfowl

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.cs
```diff
@@ -815,7 +815,7 @@ public ErrorPage()
...
....attr(\"id\");\r\n        var selected = ns(\"#\" + el.id).addClass(\"selected\").attr(\"id\");\r\n\r\n        ns(\"#\" + unselected + \"page\").hide();\r\n        ns(\"#\" + selected + \"page\").show();\r\n    }\r\n\r\n    ns(\".collapsible\").hide();\r\n    ns(\".page\").hide();\r\n    ns(\"#stackpage\").show();\r\n\r\n    ns(\"#rawExceptionButton\").click(function (event) {\r\n        var div = document.getElementById('rawException');\r\n        div.style.display = 'inline-block';\r\n        div.scrollIntoView(true);\r\n        event.preventDefault();\r\n        event.stopPropagation();\r\n\r\n        return false;\r\n    });\r\n\r\n    ns(\".frame\")\r\n        .click(function () {\r\n            frame(this);\r\n        })\r\n        .keypress(function (e) {\r\n            if (e.which === 13) {\r\n                frame(this);\r\n            }\r\n        });\r\n\r\n    ns(\"#header li\")\r\n        .click(function () {\r\n            tab(this);\r\n        })\r\n        .keypress(function (e) {\r\n            if (e.which === 13) {\r\n                tab(this);\r\n            }\r\n        });\r\n})(window);\r\n            //-->\r\n        </script>\r\n    </body>\r\n</html>\r\n");
```
davidfowl: nit: line too long

davidfowl:
:shipit:

ryanbrandenburg:
Just as a historical note, my best guess as to what happened here is that when I fixed the PageGenerator in my previous PR, the generated page was way behind what had been in the actual .cshtml and .js files, so most likely someone messed with those tabs, saw that it still worked and moved on, not realizing that it never got generated and so their code wasn't actually checked. Then I came along with my bug and generated all new stuff, but didn't click onto any of the other tabs because the one I cared about was the default and I didn't change any code that should affect them.

PR #230, opened 2015-12-15
Allow inline configure of options in UseDeveloperExceptionPage

### LINKED ISSUE #219: Allow inline configure of options in UseDeveloperExceptionPage
In most components you can use an action the set the options.
```       
             app.UseDeveloperExceptionPage(options=>options
                    {
                        SourceCodeLineCount=10;
                    });
```
instead of
```
app.UseDeveloperExceptionPage(new Microsoft.AspNet.Diagnostics.ErrorPageOptions() 
           {
             SourceCodeLineCount=10
          });
```

pranavkm:
Fixes #219

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageExtensions.cs
+++ b/src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageExtensions.cs
@@ -7,16 +7,15 @@ using Microsoft.AspNet.Diagnostics;
 namespace Microsoft.AspNet.Builder
 {
     /// <summary>
-    /// IApplicationBuilder extension methods for the ErrorPageMiddleware.
+    /// <see cref="IApplicationBuilder"/> extension methods for the <see cref="DeveloperExceptionPageMiddleware"/>.
     /// </summary>
     public static class DeveloperExceptionPageExtensions
     {
         /// <summary>
-        /// Captures synchronous and asynchronous exceptions from the pipeline and generates HTML error responses.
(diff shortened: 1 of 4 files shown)


pranavkm:
cc @Tratcher \ @Eilon

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageExtensions.cs
```diff
@@ -24,23 +23,31 @@ public static IApplicationBuilder UseDeveloperExceptionPage(this IApplicationBui
...
             {
                 throw new ArgumentNullException(nameof(builder));
             }
 
+            if (configureOptions == null)
+            {
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            var options = new ErrorPageOptions();
```
Tratcher: Did options class never renamed to the middleware?

Tratcher:
:shipit:

Eilon:
:shipit:

src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageExtensions.cs
pranavkm: Should we rename these? Using this overload is already broken for end users, so we might make the most of it and change the type name to match the middleware name. Thoughts?

Tratcher: yes

PR #236, opened 2015-12-30
Use default encoders all the time

### LINKED ISSUE #234: Diagnostics should register encoders in startup or use default values all the time
See here - https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs#L69 where the body of the extension method is:

```C#
public static HtmlEncoder GetHtmlEncoder(this IServiceProvider serviceProvider)
{
    return (HtmlEncoder)serviceProvider?.GetService(typeof(HtmlEncoder)) ?? HtmlEncoder.Default;
}
```

We should either require registering encoders in the service container or use `*Encoder.Default` all the time. I'd prefer the latter since these are diagnostic pages and are not meant for production views.

pranavkm:
Fixes #234

### DIFF
--- a/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
+++ b/src/Microsoft.AspNet.Diagnostics/Views/BaseView.cs
@@ -12,7 +12,6 @@ using System.Text;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
-using Microsoft.Extensions.WebEncoders;
 
 namespace Microsoft.AspNet.Diagnostics.Views
 {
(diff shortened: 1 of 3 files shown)


pranavkm:
cc @Eilon \ @ryanbrandenburg

Eilon:
:shipit: and I'm amazed there was a specific test for this...

ryanbrandenburg:
Seems alright. was the existing test you removed mocking to much to be useful or something?

davidfowl:
Yes! Can we also remove `Get*Encoder`

pranavkm:
Yeah, that was the plan @davidfowl

pranavkm:
@ryanbrandenburg it was verifying the scenario we removed. We just use the default encoders and not the one specified in the service container.

Issue #184, opened 2015-10-02
Consider moving DatabaseErrorPageOptions to Microsoft.AspNet.Builder namespace


Eilon:
There's a general pattern where types that are used *only* for configuring middleware also end up in the `Microsoft.AspNet.Builder` namespace, alongside the various `app.UseXyz(...)` extension methods.

The `DatabaseErrorPageOptions` is used only in middleware setup (and effectively internally elsewhere), so it should move to the `Microsoft.AspNet.Builder` namespace.

The end result is that fewer namespaces need to be imported in a typical `Startup.cs` file.

Eilon:
Just discussed another option with @Tratcher that both of us like:

Instead of passing in a particular type, the `UseXyz()` method to add the middleware could use a builder pattern and take in options that are configured inline, removing the need to reference any type by name at all, e.g.:

```c#
app.UseDatabaseErrorPage(options => {
    options.ShowExceptionDetails = true;
    options.ListMigrations = true;
});
```

Or if that's too verbose, could add a helper:

```c#
app.UseDatabaseErrorPage(options => {
    options.EnableAll(); // just sets all the properties to true
});
```

This would be even *more* inline with how other middleware works, as seen here: https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Startup.cs#L161-L218

Eilon:
@rynowak FYI

rowanmiller:
Agreed on the builder pattern. I working on #190 at the moment so I will enable this pattern at the same time.

Eilon:
@rowanmiller cool.

rowanmiller:
Fixed in https://github.com/aspnet/Diagnostics/commit/1c26436041648a36dc6c7ff4fc346fb4c2411de3

Eilon:
@rowanmiller remember to mark completed bugs as `3 - Done`.

rowanmiller:
@Eilon done... I mean `3 - Done`

Eilon:
`4 - Thanks`!

PR #239, opened 2016-01-07
Updating to new options pattern

JunTaoLuo:
https://github.com/aspnet/Options/issues/105


src/Microsoft.AspNet.Diagnostics.Entity/DatabaseErrorPageExtensions.cs
```diff
@@ -27,38 +26,9 @@ public static IApplicationBuilder UseDatabaseErrorPage(this IApplicationBuilder
...
-
-            app = app.UseMiddleware<DatabaseErrorPageMiddleware>(options);
-
-            if (options.EnableMigrationCommands)
-            {
-                app.UseMigrationsEndPoint(o => o.Path = options.MigrationsEndPointPath);
-            }
-
-            return app;
+            options.EnableAll();
```
Tratcher: This extension is inconsistent in that it doesn't pull options from DI. However, if we change it you'll loose the simple behavior. Could it be replaced `app.UseDatabaseErrorPage(new DatabaseErrorPageOptions().EnableAll());`?

Tratcher: @davidfowl

JunTaoLuo: @natemcmaster Since this will impact EF, do you know anyone else who might be interested in this change?

natemcmaster: How does this impact EF?

davidfowl: Yes we need to move this elsewhere. Is this what we want to be the default?

/cc @rowanmiller

rowanmiller: Whatever we do here should just be consistent with the developer exception page. I assume the developer exception page defaults to showing everything.

Tratcher: The DeveloperExceptionPage no longer has enable/disable options. https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageOptions.cs. It assumes that you only put it in the pipeline in Development mode.

rowanmiller: That makes sense, we probably still want folks to be able to customize the migrations end point path. This feels like a step backwards in usability though... the `app.UseDatabaseErrorPage(options => ...)` builder pattern was implemented per #184 from @Eilon and is feels like a step backwards for folks to have to create instances of an options object.

src/Microsoft.AspNet.Diagnostics.Entity/DatabaseErrorPageOptions.cs
```diff
@@ -11,28 +11,10 @@ namespace Microsoft.AspNet.Diagnostics.Entity
...
-        /// </summary>
-        public virtual bool EnableMigrationCommands { get; set; }
-
-        /// <summary>
         /// Gets or sets the path that <see cref="MigrationsEndPointMiddleware"/> will listen
         /// for requests to execute migrations commands. The middleware is only registered if
         /// <see cref="EnableMigrationCommands"/> is set to true.
         /// </summary>
-        public virtual PathString MigrationsEndPointPath { get; set; }
+        public virtual PathString MigrationsEndPointPath { get; set; } = MigrationsEndPointOptions.DefaultPath;
```
JunTaoLuo: There is no point in setting an empty `MigrationsEndPointPath` since that will just crash. We are only using this to pass the option along to `MigrationsEndPoint` so we might as well use the same default.

JunTaoLuo:
:up::date:

Tratcher:
:+1: @rowanmiller ?

src/Microsoft.AspNet.Diagnostics.Entity/DatabaseErrorPageExtensions.cs
Eilon: We would like all the middlesware to follow a consistent pattern, if possible. There have certainly been a number of different patterns in use up until recently.

rowanmiller: I really liked the builder pattern that you suggested and think we should have adopted that everywhere. I don't really care that much... just feels like we went backwards in terms of user-experience.

PR #246, opened 2016-02-03
Set the server manually

hishamco:


JunTaoLuo:
thanks @hishamco!

samples/StatusCodePagesSample/hosting.json
```diff
@@ -1,4 +0,0 @@
-{
-  "server": "Microsoft.AspNetCore.Server.Kestrel",
-  "server.urls": "http://localhost:5001/base"
```
Eilon: This one was using a different port. Does that matter for any tests or anything? @Tratcher do you remember if this was deliberate?

Tratcher: I don't recall. There shouldn't be anything special about this sample though.

Eilon: Ok nevermind then.

JunTaoLuo: I tested the samples before merging and there was no particular reason for that specific port.

PR #254, opened 2016-02-29
Remove IOptions from Diagnostics

HaoK:


src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageOptions.cs
```diff
@@ -12,6 +14,26 @@ namespace Microsoft.AspNetCore.Builder
...
+        /// </summary>
+        public DatabaseErrorPageOptions() { }
+
+        /// <summary>
+        /// Constructs and configures a new DatabaseErrorPageOptions.
+        /// </summary>
+        /// <param name="configureOptions">The configuration actions that will be applied.</param>
+        public DatabaseErrorPageOptions(IEnumerable<IConfigureOptions<DatabaseErrorPageOptions>> configureOptions) : this()
+        {
+            if (configureOptions != null)
```
lodejard: would this throw argument null exception now, instead of being an `if {...}` safety?

HaoK: Good catch, that seems better, I'll update

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/MigrationsEndPointOptions.cs
```diff
@@ -12,6 +14,31 @@ namespace Microsoft.AspNetCore.Builder
...
+        /// </summary>
+        /// <param name="configureOptions">The configuration actions that will be applied.</param>
+        public MigrationsEndPointOptions(IEnumerable<IConfigureOptions<MigrationsEndPointOptions>> configureOptions) : this()
+        {
+            if (configureOptions == null)
+            {
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            if (configureOptions != null)
```
lodejard: don't need the second if anymore :)

HaoK: So much blind copy paste fail :/  Thanks

HaoK: (This is why I hate boilerplate code fyi :)

davidfowl: Make a shared source helper :stuck_out_tongue_closed_eyes:

lodejard: Metaprogramming! Oh - wait... :3

HaoK: How about this sugar?
```C#
public static class Options {
   public static void Configure<T>(T instance, IEnumerable<IConfigureOptions<T>> configureOptions);
}

public MyOptions {
   public MyOptions() { }
   public MyOptions(IEnumerable<IConfigureOptions<MyOptions>> configureOptions) : this() {
      Options.Configure(this, configureOptions);
   }
}
```

HaoK: Tried it out, maybe is better as an extension on the IEnumerable instead:

```C#
public static class Options {
   public static void Configure<T>(this IEnumerable<IConfigureOptions<T>> configureOptions, T instance);
}

public MyOptions {
   public MyOptions() { }
   public MyOptions(IEnumerable<IConfigureOptions<MyOptions>> configureOptions) : this() {
      configureOptions.Configure(this);
   }
}
```

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageOptions.cs
```diff
@@ -12,6 +14,28 @@ namespace Microsoft.AspNetCore.Builder
...
         /// <summary>
+        /// Constructs a new DatabaseErrorPageOptions.
+        /// </summary>
+        public DatabaseErrorPageOptions() { }
+
+        /// <summary>
+        /// Constructs and configures a new DatabaseErrorPageOptions.
+        /// </summary>
+        /// <param name="configureOptions">The configuration actions that will be applied.</param>
+        public DatabaseErrorPageOptions(IEnumerable<IConfigureOptions<DatabaseErrorPageOptions>> configureOptions) : this()
```
javiercn: I'm failing to understand the new pattern. Do we need to register the custom Options classes on the DI container as a singleton? Should other option types like ElmOptions have a similar constructor like this one?

HaoK: Hold off on making any changes for a bit if you can we are still in the final iterations of the new pattern. I will update all the repos in the next few days so we should soon be in a new consistent state for options. 

But the basic general idea is there should be a default constructor that any defaults. And one "options" ctor that takes the IConfigureOptions and applies them.

Issue #257, opened 2016-03-11
Fix diagnostics packages dependencies


Eilon:
Both the EF database error middleware and the ELM middleware depend on the concrete Diagnostics package. They should instead depend on various abstraction packages because there is no actual relationship between these various packages.

Eilon:
https://github.com/aspnet/Diagnostics/commit/0f74187a81802a6b81366f1ea5b4d6d97cdd11e8

PR #258, opened 2016-03-11
Fix package dependencies

Eilon:
Removed some extraneous dependencies and instead made all the
diagnostics packages be independent packages that happen to share some
concepts/source code.

The common Razor base class code was moved to a shared source file. Then
the Diagnostics page, ELM page, and EF database page all use that. Also
updated the page generator to use the new shared code.

See #257

@rowanmiller @ajcvickers @divega @davidfowl - this is a fix from the meeting earlier today.

### DIFF
--- a/NuGetPackageVerifier.json
+++ b/NuGetPackageVerifier.json
@@ -14,6 +14,7 @@
     "adx-nonshipping": {
         "rules": [],
         "packages": {
+            "Microsoft.AspNetCore.DiagnosticsViewPage.Sources": { },
             "PageGenerator": { }
         }
     },
--- a/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/project.json
+++ b/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/project.json
@@ -20,16 +20,18 @@
     "xmlDoc": true
   },
   "dependencies": {
-    "Microsoft.AspNetCore.Diagnostics": "1.0.0-*",
+    "Microsoft.AspNetCore.DiagnosticsViewPage.Sources": {
+      "type": "build",
+      "version": "1.0.0-*"
+    },
+    "Microsoft.AspNetCore.Http.Abstractions": "1.0.0-*",
     "Microsoft.EntityFrameworkCore.Relational": "1.0.0-*"
   },
-  "compile": [
-    "..\\Shared\\*.cs"
-  ],
   "frameworks": {
     "net451": {
       "frameworkAssemblies": {
-        "System.Configuration": ""
+        "System.Configuration": "",
+        "System.IO": ""
       }
     },
     "netstandard1.3": {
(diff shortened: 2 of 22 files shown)


src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/project.json
```diff
@@ -20,12 +20,13 @@
...
-    "Microsoft.AspNetCore.Diagnostics": "1.0.0-*",
+    "Microsoft.AspNetCore.DiagnosticsViewPage.Sources": {
+      "type": "build",
+      "version": "1.0.0-*"
+    },
+    "Microsoft.AspNetCore.Http.Abstractions": "1.0.0-*",
     "Microsoft.EntityFrameworkCore.Relational": "1.0.0-*"
   },
-  "compile": [
-    "..\\Shared\\*.cs"
```
Eilon: This was apparently not used at all.

src/Microsoft.AspNetCore.DiagnosticsViewPage.Sources/project.json
```diff
@@ -0,0 +1,8 @@
+{
+  "version": "1.0.0-*",
+  "shared": "*.cs",
+  "dependencies": {},
+  "frameworks": {
```
BrennanConroy: Should this have net451 as well?

Eilon: It doesn't seem to actually matter because it's a sources package. None of the code compiles in the context of this project. You could even have a compiler error and it doesn't care.

I just copied what some of the other shared source packages do.

Eilon:
@kichalla can you review?

NuGetPackageVerifier.json
```diff
@@ -14,6 +14,7 @@
     "adx-nonshipping": {
         "rules": [],
         "packages": {
+            "Microsoft.AspNetCore.DiagnosticsViewPage.Sources": { },
```
kichalla: Nice. I suppose you are planning to update the `.csv` file too.

kichalla:
:shipit: !

NuGetPackageVerifier.json
Eilon: Oh right! Thanks for the reminder.

Eilon: @kichalla added it here: https://github.com/aspnet/specs/commit/c0533ba951992e697f9a9afc8f39102ad13d3bd4

muratg: @Eilon `packages.csv`?  This was moved to Coherence repo (to make downloading the file easier).  Please update both Coherence-Signed and Coherence copies. We'll delete the one in Coherence-Signed soon.

Eilon:
Thanks for the review!

PR #259, opened 2016-03-11
Target dnx451

bricelam:
This reverts e57e1772244b340bc281d6e2205db5848ae55c24 for DNX


bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

PR #260, opened 2016-03-18
Addition of InstanceId to the DS messages being published.

avanderhoorn:
As consumers of the middleware diagnostics source events we find ourselves in situation where we are needing to correlate being/end messages together in a reliable way. We have attempted to do this on the consumer side without any help from the publisher but to date have come up short. 

The following outlines the cases in question:
```
- CoresMiddleware Start
    - LoggingMiddleware Start
        - StaticFileMiddleware Start
            - FizBuzMiddleware Start
                - StaticFileMiddleware Start
                     ....
                - StaticFileMiddleware End
            - FizBuzMiddleware End
        - StaticFileMiddleware End
    - LoggingMiddleware End
- CoresMiddleware End
```
As you can see, we have a fairly standard pipeline with the only semi unusual thing is that `StaticFileMiddleware` is regsitered twice in the pipeline.

Looking at the above, its quite simple to say that we could use "execution order" as the means by which we correlate begin/end pairs, but this doesn't work all that well when/if we have gaps in the messages. In theory the publisher can "guarantee" that it will publish these pairs, but on the consumer side, we aren't guarantee that all these messages will get through our system in an ACID compliant way. 

Specifically, if the client that is ultimately displaying or displaying the data ends up with the following messages, how does it know which `StaticFileMiddleware being` the one instance of the `StaticFileMiddleware End` we have:
```
Ordinal  -|-  Instance
1   -|-  CoresMiddleware Start
3   -|-  LoggingMiddleware Start
8   -|-  StaticFileMiddleware Start
9   -|-  FizBuzMiddleware Start
12  -|-  StaticFileMiddleware Start
17  -|-  StaticFileMiddleware End
21  -|-  FizBuzMiddleware End
<<< Missing
43  -|-  LoggingMiddleware End
49  -|-  CoresMiddleware End
```

Now, looking at the above we could say we use the `name` and `ordinal` to do that match on, but that makes a lot of assumptions around that uniqueness of the names and is fraught with edge cases. Additionally, as mentioned, we attempted to track/correlate this on the point of consumption, but given the consumer model and the context we live in, there is no real reliable way of us doing this.

All this lead us to the point where we believe the publisher is in the best position to positively identify being/end pairs with absolute reliability and makes the consumers life 1000x easier in terms of positively identifying begin/end pairs.

This `id` has been placed at the instance level, as we only need to be able to positively/uniquely identify pairs within the boundaries of a give request, and even enables use cases where we could correlate middleware instances across requests (though this is a low priority use case). Additionally, by putting this at the instance level, we are limiting the number of ids we need to generated to the fix number of pieces of middleware registered, instead of the number of times middleware runs.

// @Tratcher @lodejard

### DIFF
--- a/src/Microsoft.AspNetCore.MiddlewareAnalysis/AnalysisMiddleware.cs
+++ b/src/Microsoft.AspNetCore.MiddlewareAnalysis/AnalysisMiddleware.cs
@@ -10,6 +10,7 @@ namespace Microsoft.AspNetCore.MiddlewareAnalysis
 {
     public class AnalysisMiddleware
     {
+        private readonly Guid _instanceId = Guid.NewGuid();
         private readonly RequestDelegate _next;
         private readonly DiagnosticSource _diagnostics;
         private readonly string _middlewareName;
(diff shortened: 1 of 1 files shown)


Eilon:
This seems reasonable.

Tratcher:
Rebased and merged.

PR #261, opened 2016-03-23
Add timestamp and duration to Diagnostics Source events

avanderhoorn:
Given the precedence of what we have here https://github.com/aspnet/Hosting/pull/543 I'm adding timestamp data to these events - this ensures that everyone is working from the same base data when talking timings. 

In addition, I'm adding `duration` to the `finish` events so that consumers only have to subscribe to `finish` events in order to determine what the duration for a given piece of middleware is. This change is localized to the scope of code where we know that we have listeners interested in the result and hence isn't on the primary execution path.

### DIFF
--- a/src/Microsoft.AspNetCore.MiddlewareAnalysis/AnalysisMiddleware.cs
+++ b/src/Microsoft.AspNetCore.MiddlewareAnalysis/AnalysisMiddleware.cs
@@ -28,10 +28,9 @@ namespace Microsoft.AspNetCore.MiddlewareAnalysis
 
         public async Task Invoke(HttpContext httpContext)
         {
-            var startTimestamp = 0L;
+            var startTimestamp = Stopwatch.GetTimestamp();
             if (_diagnostics.IsEnabled("Microsoft.AspNetCore.MiddlewareAnalysis.MiddlewareStarting"))
             {
-                startTimestamp = Stopwatch.GetTimestamp();
                 _diagnostics.Write(
                     "Microsoft.AspNetCore.MiddlewareAnalysis.MiddlewareStarting",
                     new
--- a/src/Microsoft.AspNetCore.MiddlewareAnalysis/AnalysisMiddleware.cs
+++ b/src/Microsoft.AspNetCore.MiddlewareAnalysis/AnalysisMiddleware.cs
@@ -28,9 +28,11 @@ namespace Microsoft.AspNetCore.MiddlewareAnalysis
 
         public async Task Invoke(HttpContext httpContext)
         {
+            var startTimestamp = 0L;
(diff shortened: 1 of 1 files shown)


Tratcher:
@lodejard @rynowak @Eilon

src/Microsoft.AspNetCore.MiddlewareAnalysis/AnalysisMiddleware.cs
```diff
@@ -28,9 +28,11 @@ public AnalysisMiddleware(RequestDelegate next, DiagnosticSource diagnosticSourc
 
         public async Task Invoke(HttpContext httpContext)
         {
+            var startTimestamp = 0L;
             if (_diagnostics.IsEnabled("Microsoft.AspNetCore.MiddlewareAnalysis.MiddlewareStarting"))
             {
-                _diagnostics.Write("Microsoft.AspNetCore.MiddlewareAnalysis.MiddlewareStarting", new { name = _middlewareName, httpContext = httpContext, instanceId = _instanceId });
+                startTimestamp = Stopwatch.GetTimestamp();
+                _diagnostics.Write("Microsoft.AspNetCore.MiddlewareAnalysis.MiddlewareStarting", new { name = _middlewareName, httpContext = httpContext, instanceId = _instanceId, timestamp = startTimestamp });
```
Eilon: These lines are getting a tad long - can you wrap them like:

```c#
_diagnostics.Write(
    "Microsoft.AspNetCore.MiddlewareAnalysis.MiddlewareStarting",
        new
        {
            name = _middlewareName,
            httpContext = httpContext,
            instanceId = _instanceId,
            timestamp = startTimestamp, // and include this trailing comma :)
        });
```

Eilon:
@kichalla please review & merge when you have a chance.

kichalla:
> In addition, I'm adding duration to the finish events so that consumers only have to subscribe to finish events in order to determine what the duration for a given piece of middleware is.

If users subscribe only to the `finish` event, then the `startTimestamp` would be zero causing the duration to be incorrect. Should we always capture `startTimestamp` then? (but yeah i see you mention about primary execution path)

avanderhoorn:
I agree with the case you mentioned... As you mentioned, one approach is to

avanderhoorn:
Whoops didn't meant to close... reopend.

avanderhoorn:
Ok, I've updated things to fix the line breaks. Additionally, I've updated things so that the `startTimestamp` is always captured. If we are worried about always capturing that, we can update things to only capture it if either the `MiddlewareFinished` or `MiddlewareException` are being listened to, but that check could be more expensive than just capturing the value.

kichalla:
:shipit:

Eilon:
Ah yeah I like this fix - I'd like the data to always be accurate, when available.

kichalla:
f387d04af12fcfd5fc79d3c09c36df3cfbdefc41

PR #270, opened 2016-04-18
Fix Apply Migrations button

rowanmiller:
During all the changes to how options are handled etc. we ended up where
app.UseDatabaseErrorPage() would no longer register the migrations
middleware that allows the "Apply Migrations" button to work.

Resolves aspnet/EntityFramework#4742

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageExtensions.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageExtensions.cs
@@ -14,8 +14,7 @@ namespace Microsoft.AspNetCore.Builder
     {
         /// <summary>
         /// Captures synchronous and asynchronous database related exceptions from the pipeline that may be resolved using Entity Framework
-        /// migrations. When these exceptions occur an HTML response with details of possible actions to resolve the issue is generated. The
-        /// options for the middleware are set to display the maximum amount of information available.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageExtensions.cs
```diff
@@ -26,7 +25,7 @@ public static IApplicationBuilder UseDatabaseErrorPage(this IApplicationBuilder
                 throw new ArgumentNullException(nameof(app));
             }
 
-            return app.UseMiddleware<DatabaseErrorPageMiddleware>();
+            return app.UseDatabaseErrorPage(new DatabaseErrorPageOptions());
```
Eilon: Curious - why didn't the old code work?

natemcmaster: It looks like `MigrationsEndPointMiddleware` was never wired up on this code path. The error page would show up, but the endpoint to apply migrations was never added.

natemcmaster:
:shipit:

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageExtensions.cs
rowanmiller: Yep, the migrations end point is a separate piece of middleware. The overload we now delegate to adds both pieces of middleware.

PR #273, opened 2016-04-25
Update web.config and add publish tool

BrennanConroy:
@JunTaoLuo @Eilon @Tratcher


muratg:
:shipit: once you get approval :)

Issue #268, opened 2016-04-13
UseStatusCodePagesWithReExecute should be able to take a query string


ryanbrandenburg:
Currently if you pass a query string on pathFormat like "/Errors/Index?id={0}" it won't work, and will prevent the route from resolving (because we're setting the formatted string directly onto HttpContext.Request.Path, where query string should already be broken off).

We should consider adding the capability to pass query strings to this function, perhaps as a "queryFormat" parameter.

CC @Eilon, @Tratcher

ryanbrandenburg:
Doing this means we'll also have to do #267 to ensure the original query string is preserved.

drewlsvern:
What value is queryFormat supposed to be? I can't find any documentation on this. 

Any help would be appreciated.

Tratcher:
@drewlsvern see this sample in the test: https://github.com/aspnet/Diagnostics/pull/276/files#diff-cdde232dfc093c600857f42a907fd478R244

PR #276, opened 2016-04-26
Allow query strings on UseStatusCodePagesWithReExecute

### LINKED ISSUE #267: UseStatusCodePagesWithReExecute shouldn't pass through the QueryString from the original request
[This](https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodePagesExtensions.cs#L162) method currently will pass the Query string from the original request to the re-executed request, even though the path has changed. @Eilon  and I were thinking that it should probably be cleared and restored at the same time the original path is restored.

ryanbrandenburg:
Fixes #267, #268 

CC @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodePagesExtensions.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodePagesExtensions.cs
@@ -158,8 +158,12 @@ namespace Microsoft.AspNetCore.Builder
         /// </summary>
         /// <param name="app"></param>
         /// <param name="pathFormat"></param>
+        /// <param name="queryFormat"></param>
         /// <returns></returns>
-        public static IApplicationBuilder UseStatusCodePagesWithReExecute(this IApplicationBuilder app, string pathFormat)
+        public static IApplicationBuilder UseStatusCodePagesWithReExecute(
+            this IApplicationBuilder app,
+            string pathFormat,
+            string queryFormat = null)
         {
             if (app == null)
             {
@@ -168,23 +172,31 @@ namespace Microsoft.AspNetCore.Builder
 
             return app.UseStatusCodePages(async context =>
             {
-                var newPath = new PathString(string.Format(CultureInfo.InvariantCulture, pathFormat, context.HttpContext.Response.StatusCode));
+                var newPath = new PathString(
+                    string.Format(CultureInfo.InvariantCulture, pathFormat, context.HttpContext.Response.StatusCode));
+                var formatedQueryString = queryFormat == null ? null :
+                    string.Format(CultureInfo.InvariantCulture, queryFormat, context.HttpContext.Response.StatusCode);
+                var newQueryString = queryFormat == null ? QueryString.Empty : new QueryString(formatedQueryString);
 
                 var originalPath = context.HttpContext.Request.Path;
+                var originalQueryString = context.HttpContext.Request.QueryString;
                 // Store the original paths so the app can check it.
                 context.HttpContext.Features.Set<IStatusCodeReExecuteFeature>(new StatusCodeReExecuteFeature()
                 {
                     OriginalPathBase = context.HttpContext.Request.PathBase.Value,
                     OriginalPath = originalPath.Value,
+                    OriginalQueryString = originalQueryString.HasValue ? originalQueryString.Value : null,
                 });
 
                 context.HttpContext.Request.Path = newPath;
+                context.HttpContext.Request.QueryString = newQueryString;
                 try
                 {
                     await context.Next(context.HttpContext);
                 }
                 finally
                 {
+                    context.HttpContext.Request.QueryString = originalQueryString;
                     context.HttpContext.Request.Path = originalPath;
                     context.HttpContext.Features.Set<IStatusCodeReExecuteFeature>(null);
                 }
--- a/src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodeReExecuteFeature.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodeReExecuteFeature.cs
@@ -8,5 +8,7 @@ namespace Microsoft.AspNetCore.Diagnostics
         public string OriginalPath { get; set; }
 
         public string OriginalPathBase { get; set; }
+
+        public string OriginalQueryString { get; set; }
     }
 }
--- a/test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
+++ b/test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
@@ -176,6 +176,106 @@ namespace Microsoft.AspNetCore.Diagnostics
         }
 
         [Fact]
+        public async Task Redirect_StatusPage()
+        {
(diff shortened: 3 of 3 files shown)


src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodePagesExtensions.cs
```diff
@@ -169,22 +173,29 @@ public static IApplicationBuilder UseStatusCodePagesWithReExecute(this IApplicat
...
 
                 context.HttpContext.Request.Path = newPath;
+                context.HttpContext.Request.QueryString = newQueryString;
                 try
                 {
                     await context.Next(context.HttpContext);
                 }
                 finally
                 {
+                    // context.HttpContext.Request.QueryString = originalQueryString;
```
Tratcher: ?

src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodeReExecuteFeature.cs
```diff
@@ -1,12 +1,16 @@
...
+
 namespace Microsoft.AspNetCore.Diagnostics
 {
     public class StatusCodeReExecuteFeature : IStatusCodeReExecuteFeature
     {
         public string OriginalPath { get; set; }
 
         public string OriginalPathBase { get; set; }
+
+        public QueryString OriginalQueryString { get; set; }
```
Tratcher: Why a QueryString rather than just a string like the paths?

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
```diff
@@ -176,6 +176,48 @@ public class ExceptionHandlerTest
         }
 
         [Fact]
+        public async Task Reexecute_RequestWithQueryString()
```
Tratcher: Add a similar test for UseStatusCodePagesWithRedirects

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
```diff
@@ -176,6 +176,48 @@ public class ExceptionHandlerTest
...
+            var builder = new WebHostBuilder()
+                .Configure(app =>
+                {
+                    app.UseStatusCodePagesWithReExecute("/errorPage", "?id={0}");
+
+                    app.Map(destination, (innerAppBuilder) =>
+                    {
+                        innerAppBuilder.Run(async (httpContext) =>
+                        {
+                            await Task.Delay(1);
```
muratg: Nit: use TimeSpan to make the delay explicit.

Tratcher: Why is there a delay here?

ryanbrandenburg: Just so that the delegate was async and VS wouldn't complain, I've come up with a more useful thing to do there though.

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
ryanbrandenburg: Am I missing some inter-op between `UseStatusCodePagesWithReExecute` and `UseStatusCodePagesWithRedirects` or are you just saying that in general it's missing a test and I may as well add it in this PR?

src/Microsoft.AspNetCore.Diagnostics/StatusCodePage/StatusCodePagesExtensions.cs
ryanbrandenburg: This was me checking an assumption and I forgot to revert it. Will fix.

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
```diff
@@ -176,6 +176,48 @@ public class ExceptionHandlerTest
...
+                        throw new InvalidOperationException("Invalid input provided.");
+                    });
+                });
+
+            using (var server = new TestServer(builder))
+            {
+                var client = server.CreateClient();
+                var response = await client.GetAsync(destination);
+                Assert.Equal($"?id={expectedStatusCode}", await response.Content.ReadAsStringAsync());
+                Assert.Equal("", response.RequestMessage.RequestUri.Query);
```
ryanbrandenburg: This apparently won't actually be populated if the query string isn't reset after the ReExecution, is there a better way to check that the query string was properly re-set for the rest of the pipeline?

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
Tratcher: It's just missing a test for the same scenario.

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
Tratcher: Put a Use middleware at the front of the app that verifies the query after next.

ryanbrandenburg:
:up::date:

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
```diff
@@ -388,5 +481,26 @@ public class ExceptionHandlerTest
             Assert.NotNull(listener.DiagnosticHandledException?.HttpContext);
             Assert.NotNull(listener.DiagnosticHandledException?.Exception);
         }
+
+        public class InspectAfterReExecuteMiddleware
+        {
+            public const string QueryStringDifferenceHeader = "QUERYSTRINGDIFFERENT";
```
Tratcher: ?

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
```diff
@@ -388,5 +481,26 @@ public class ExceptionHandlerTest
...
+
+            private readonly RequestDelegate _next;
+
+            public InspectAfterReExecuteMiddleware(RequestDelegate next)
+            {
+                _next = next;
+            }
+
+            public async Task Invoke(HttpContext context)
```
Tratcher: You can inline this:
```
app.Use(async (context, next) =>
{
  var beforeNext = context.Request.QueryString;
  await _next(context);
  var afterNext = context.Request.QueryString;
  Assert.Equal(beforeNext, afterNext);
});
```

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
ryanbrandenburg: I had been trying to return the success or failure in the header but realized I could just do the assert in the Middleware, missed it in cleanup.

Tratcher:
:shipit:

Issue #293, opened 2016-05-17
Exceptions during razor views rendering don't show the source line


sebastienros:
Can repro in a new WebApp from template. 

Repro:
- Update `Index.cshtml` with `@(((string)null).ToString()`

Result:
![image](https://cloud.githubusercontent.com/assets/1165805/15341324/dcf4f06a-1c42-11e6-80c2-5b80241529c6.png)

Expected:
It should display the line number where the exception occurred.

NB:
- Looking at the `sourceTree` value the line numbers are generated in the razor source.
- This scenario should be added in release checks

pranavkm:
Seems like an issue with compiling using the `PortablePdb` switch. I asked the Roslyn folks at https://github.com/dotnet/roslyn/issues/11535 for guidance on this.

pranavkm:
The issue's been fixed for RTM - https://github.com/dotnet/coreclr/issues/3906. We're working on picking up the packages now so this is likely to get fixed soon. I filed https://github.com/aspnet/Mvc/issues/4737 to add a functional test to verify this behavior is covered.

pranavkm:
The RC3 packages fix this issue in netcoreapp, but it's broken on desktop. I guess we could generate full pdbs if the right components are present. @davidfowl, does that sound ok?

dougbu:
@pranavkm why is dotnet/coreclr#3906 considered closed if the problem is only fixed when using .NET core? Have you filed a CoreFx bug for full desktop support?

pranavkm:
@davidfowl's going to commonize the pdb reader (https://github.com/aspnet/Diagnostics/pull/300) and use it in Hosting and Diagnostics. I'll check in the test part of https://github.com/aspnet/Mvc/commit/73ddec7c67d7de15cdee91163c0d4078904274f6 once this is done to get some e2e coverage.

PR #300, opened 2016-06-01
Added support for reading portable PDBs on .NET Framework/Mono apps

davidfowl:
TODO: Add a test and verify on older mono versions

- Mono 4.4 supports portable PDB and the StackTrace API just works
- This code only kicks in if the filename is null and we're on .NET 4.5.x
- Would be good to test out a dynamic assembly though it should be handled by checking for a null assembly location.

Edit: Mono 4.2.4 crashes and burns with PPDB trying to get the stack trace information:

```C#
Stacktrace:
  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) System.Diagnostics.StackTrace.get_trace (System.Exception,int,bool) <IL 0x00027, 0xffffffff>
  at System.Diagnostics.StackTrace..ctor (System.Exception,int,bool) [0x0002e] in /private/tmp/source-mono-4.2.4/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.4/mcs/class/corlib/System.Diagnostics/StackTrace.cs:129
* Assertion at debug-mono-ppdb.c:217, condition `cols [MONO_METHODBODY_SEQ_POINTS]' not met
  at System.Diagnostics.StackTrace..ctor (System.Exception,bool) [0x00000] in /private/tmp/source-mono-4.2.4/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.4/mcs/class/corlib/System.Diagnostics/StackTrace.cs:113
* Assertion at debug-mono-ppdb.c:217, condition `cols [MONO_METHODBODY_SEQ_POINTS]' not met
```

Need to see if we can detect this and just use our APIs directly. We could always use our these APIs when running on mono if the pdb is a portable pdb. We could also fallback to parsing the stack trace string again (yuck!)

/cc @tmat @pranavkm 

#293

### DIFF
--- a/test/Microsoft.AspNetCore.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
+++ b/test/Microsoft.AspNetCore.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
@@ -22,7 +22,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
 using Xunit;
-using StackFrame = Microsoft.AspNetCore.Diagnostics.Views.StackFrame;
+using StackFrameModel = Microsoft.AspNetCore.Diagnostics.Views.StackFrameModel;
 using Moq;
 
 namespace Microsoft.AspNetCore.Diagnostics
@@ -273,7 +273,7 @@ namespace Microsoft.AspNetCore.Diagnostics
         {
             // Arrange
             var middleware = GetErrorPageMiddleware();
-            var stackFrame = new StackFrame();
+            var stackFrame = new StackFrameModel();
 
             // Act
             middleware.ReadFrameContent(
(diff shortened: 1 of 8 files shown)


test/Microsoft.AspNetCore.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
```diff
@@ -22,7 +22,7 @@
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
 using Xunit;
-using StackFrame = Microsoft.AspNetCore.Diagnostics.Views.StackFrame;
+using StackFrameModel = Microsoft.AspNetCore.Diagnostics.Views.StackFrameModel;
```
davidfowl: Will remove this

Tratcher:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Startup/StartupExceptionPage.cs too?

davidfowl:
Yap, will do that after I flesh out the mono bugginess. Also we don't have a shared component for getting the portable pdb stack trace (yet)

PR #311, opened 2016-06-30
Exception Page showing only method names in the callstack

### LINKED ISSUE #304: Exception Page showing only method names in the callstack
image

Using:
dotnet 1.0.0-preview2-003024
Microsoft.Extensions.Diagnostics 1.0.0-rc3-21343

pranavkm:
Fixes #304

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
@@ -189,7 +189,7 @@ namespace Microsoft.AspNetCore.Diagnostics
                 {
                     Error = scan,
                     StackFrames = StackTraceHelper.GetFrames(ex)
-                        .Select(frame => GetStackFrame(frame.Method, frame.FilePath, frame.LineNumber))
+                        .Select(frame => GetStackFrame(frame.MethodDisplayInfo.ToString(), frame.FilePath, frame.LineNumber))
                 };
             };
         }
--- a/src/Microsoft.AspNetCore.Diagnostics/project.json
+++ b/src/Microsoft.AspNetCore.Diagnostics/project.json
@@ -35,6 +35,10 @@
       "type": "build",
       "version": "1.1.0-*"
     },
+    "Microsoft.Extensions.TypeNameHelper.Sources": {
+      "type": "build",
+      "version": "1.1.0-*"
+    },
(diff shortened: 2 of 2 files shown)


pranavkm:
@rynowak \ @halter73 

image

pranavkm:
Ping

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/StackFrameInfo.cs
```diff
@@ -1,54 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
BrennanConroy: Are you allowed to remove this? Its public so "breaking change"

pranavkm: cc @Eilon

halter73: Ugh. With all the hacks, erm... workarounds, we're going to have to do to maintain binary compat, keeping a class around that no one will ever use (or indeed *has* ever used looking at the git history) is pretty minor I guess.

Presumably our tools for checking binary compat will yell at us later if we remove it now.

@Eilon Do we want to file issues for all the easy-to-make breaking changes we want to make in the next major release? Should we create a tag for this?

Eilon: I'll work with folks to come up with official guidance on this. We definitely can't just remove this. But we can perhaps both document it as being obsolete, as well as mark it as `[Obsolete]` (*warning*, not *error*), to alert users. And then file a bug for 2.0.0 to delete it (we could actually go create that milestone if needed).

pranavkm: I'll leave this type in. We should've put all the view generation code in an internal namespace - none of it is very useful.

pranavkm:
:up: :date:

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/StackFrameInfo.cs
pranavkm: I'll do the `Obsolete` attribute in a follow up. It's not particularly related to this change.

Eilon:
:shipit:

PR #312, opened 2016-07-06
Add support for showing failure summaries.

pranavkm:
cc @ajaybhargavb \ @Eilon


pranavkm:
This is what it looks like:

image

The message is repeated twice once for code with line mappings and once for the generated code which doesn't exist in the page. It's not super terrible to look at and avoids creating extra interfaces. Thoughts?

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.cshtml
```diff
@@ -34,6 +34,10 @@
                         <div class="titleerror">@fileName</div>
                     }
                 }
+                @if (!string.IsNullOrEmpty(errorDetail.ErrorMessage))
+                {
+                    <details>@errorDetail.ErrorMessage</details>
```
Eilon: I do not think this tag does what you think it does.

https://www.w3.org/wiki/HTML/Elements/details

> HTML Attributes
open = boolean
If present, it indicates that the details are to be shown to the user. If the attribute is absent, the details are not to be shown. [Example B] 

How about a `<div>` or `<p>`?

Eilon:
The design seems :shipit: 

Is it possible to add a test for this?

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.cshtml
pranavkm: Sure.

pranavkm:
We have integration tests for Diagnostics in Mvc - https://github.com/aspnet/Mvc/pull/4968 (including this feature). I don't think Diagnostics has any independent tests of its own around compilation

Eilon:
@pranavkm re: tests -- got it.

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.cshtml
dougbu: @Eilon doesn't this create a control that's currently closed but can be opened to show the details? `open="open"` would create a control that's open but can be closed.

@pranavkm which browsers support `<details>`?

pranavkm: Chrome v51 and Edge seem to treat it as a regular `div` and I don't see any gestures to close it (it's visible by default).

src/Microsoft.AspNetCore.Diagnostics.Abstractions/CompilationFailure.cs
```diff
@@ -10,12 +10,44 @@ namespace Microsoft.AspNetCore.Diagnostics
     /// </summary>
     public class CompilationFailure
     {
-        public CompilationFailure(string sourceFilePath, string sourceFileContent, string compiledContent, IEnumerable<DiagnosticMessage> messages)
+        /// <summary>
+        /// Initializes a new instance of <see cref="CompilationFailure"/>.
+        /// </summary>
+        /// <param name="sourceFilePath">Path for the file that produced the compilation failure.</param>
+        /// <param name="sourceFileContent">Contents of the file being compiled.</param>
+        /// <param name="compiledContent">For templated languages (such as Asp.Net Razor), the generated content.
```
dougbu: Asp.Net _Core_ Razor?

ajaybhargavb:
:shipit:

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.cs
```diff
@@ -63,7 +63,7 @@ public CompilationErrorPage()
...
...ource .highlight li span {\r\n    color: #fff;\r\n    background: #b20000;\r\n}\r\n\r\n#stackpage .source ol.collapsible li {\r\n    color: #888;\r\n}\r\n\r\n    #stackpage .source ol.collapsible li span {\r\n        color: #606060;\r\n    }\r\n\r\n.page table {\r\n    border-collapse: separate;\r\n    border-spacing: 0;\r\n    margin: 0 0 20px;\r\n}\r\n\r\n.page th {\r\n    vertical-align: bottom;\r\n    padding: 10px 5px 5px 5px;\r\n    font-weight: 400;\r\n    color: #a0a0a0;\r\n    text-align: left;\r\n}\r\n\r\n.page td {\r\n    padding: 3px 10px;\r\n}\r\n\r\n.page th, .page td {\r\n    border-right: 1px #ddd solid;\r\n    border-bottom: 1px #ddd solid;\r\n    border-left: 1px transparent solid;\r\n    border-top: 1px transparent solid;\r\n    box-sizing: border-box;\r\n}\r\n\r\n    .page th:last-child, .page td:last-child {\r\n        border-right: 1px transparent solid;\r\n    }\r\n\r\n    .page .length {\r\n        text-align: right;\r\n    }\r\n\r\na {\r\n    color: #1ba1e2;\r\n    text-decoration: none;\r\n}\r\n\r\n    a:hover {\r\n        color: #13709e;\r\n        text-decoration: underline;\r\n    }\r\n\r\n        </style>\r\n    </head>\r\n    <body>\r\n        <h1>");
```
pranavkm: Unrelated, but should we generate these files as part of `build` rather than check it in? It's just so much noise when things change.

src/Microsoft.AspNetCore.Diagnostics.Abstractions/CompilationFailure.cs
Eilon: I'd just say Razor. (And BTW it's ASP.NET, never Asp.Net.)

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.cshtml
Eilon: http://caniuse.com/#feat=details indicate that about 76% of browsers in use today correctly support the element. Let's just use a `<div>`.

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.cs
Eilon: That could be nice. I think I fixed it up a ton a while ago so that it would actually *work* (the generator was completely broken for a while). Making it run on every build is perhaps an improvement. BUT, if we do that, I'd like to have tests that do more thorough testing of the rendered output in a variety of scenarios. Right now the risk is relatively low when nothing changes in this repo because the files are checked in. But if we re-run Razor on every build, that introduces risk because there might be some inadvertent change that we don't easily or quickly catch.

pranavkm: We could pin the version of Razor to RTM. We really don't use bleeding edge features here and it removes one more dependency edge from our build graph.

Eilon: I can live with that.

pranavkm: https://github.com/aspnet/Diagnostics/pull/313 already does this. I'll follow up with a script to run the tool as part of `precompile`.

pranavkm: Looks like using `precompile` doesn't work (See https://github.com/dotnet/cli/issues/3807) in this scenario. Doing it in our build scripts is problematic because you have to remember to run it before opening the solution. I guess I'll leave it as is :disappointed:

Issue #319, opened 2016-07-20
It is not possible to copy text to clipboard from error message on development page.


maxim-boligatov:
_From @maxim-boligatov on July 20, 2016 7:51_

When exception occurs development page is displayed in browser.  
Sometimes it significantly increases productivity when it is possible to copy error text or sample message from error page.  
For example for the following example I can just find RenderSection line in source:

```
         <!-- Styles -->
         @this.RenderSection("styles", required: false)
        </head>
        <body>
            @RenderBody()
        </body>
        </html>

```

But due to interactivity added to development page (all mouse events are handled) I cannot copy text to clipboard easy. So I have to extract it from HTML using firebug...
How can it be fixed/improved? What do you think?

_Copied from original issue: aspnet/Mvc#5048_

maxim-boligatov:
_From @pranavkm on July 20, 2016 13:46_

Could you file this bug in the [Diagnostics](https://github.com/aspnet/Diagnostics) repo?

Eilon:
I thought at the bottom of the Diagnostics page there's a "Show raw details" link that makes it easy to copy & paste?

Eilon:
Indeed, at the bottom of the developer exception page there's a "Show raw exception details" link that shows the plaintext exception message and stack trace.

If you click that link, you get text that's easy to copy and paste:

![raw-exception-details](https://cloud.githubusercontent.com/assets/202643/17000004/86b61340-4e75-11e6-9bb6-3d7f68471e01.png)

pranavkm:
I thought @maxim-boligatov was referring to the actual source code (i.e. the code from the cs \ cshtml file) rather than the stack trace?

Eilon:
Oh the source code being copyable? I'm not sure why you'd want to copy the source code... it's clearly not good source code if it threw an exception 😄

maxim-boligatov:
It is good explained above.. a search of copied string it is the fastest way to find the line which was a reason of an error (we cannot click on error and open VS with related source code... at least for now and without browser plugins).  
I think you agree that each error message will sound like "This specific feature is failed during this specific action". In case of localized resources it will be still some unique piece of source code. So I just need to search it in solution... Analyzing of stack trace will be much more slower to identify code location.

In other words due to new interactivity "features" exception page does not allow text copying.
And this interactivity is really strange decision... Do you need more space for information? Put it into 2 columns... No professional developer works on monitor with resolution less than FullHD...
[...]

PR #321, opened 2016-07-26
[Fixes #288] DeveloperExceptionPage needs better text spacing on OSX/Safari

### LINKED ISSUE #288: DeveloperExceptionPage needs better text spacing on OSX/Safari
image

kichalla:
**Compile time error:**
**Before**
compiletimeerror-before

**After**
compiletimeerror-after

**Runtime error**
**Before**
runtimeerror-before

**After**
runtimeerror-after

@Eilon @rynowak This issue repros on all browsers actually. We didn't catch it probably because we probably didn't test with exceptions which is resulting in multi-line error message(which is what view compilation error is generating)


pranavkm:
Could we review at all the styling \ formatting issues with the Diagnostics page in general rather than changing them piecemeal like this? https://github.com/aspnet/Diagnostics/issues/319 is another issue that would be useful to address.

Eilon:
@pranavkm we could, but this issue is the most pressing one, so it'd be good to just get that fix out. The other bug is a bit more work.

Eilon:
:shipit:

kichalla:
e3e092ec6030b0d8e336ec63c8040ac46ba29009

PR #322, opened 2016-07-28
[Fixes #319] It is not possible to copy text to clipboard from error …

### LINKED ISSUE #319: It is not possible to copy text to clipboard from error message on development page.

kichalla:
…message on development page.

@Eilon @pranavkm

[Internal Server Error.zip](https://github.com/aspnet/Diagnostics/files/389199/Internal.Server.Error.zip)


kichalla:
You can download the attached zip file which contains a single html file to see how the changes look like

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.cshtml
```diff
@@ -86,16 +86,21 @@
 
         <div id="stackpage" class="page">
             <ul>
-                @{ int tabIndex = 6; }
+                @{ int exceptionCount = 0; int stackFrameCount = 0; string exceptionDetailId = ""; string frameId = ""; }
```
Eilon: wut? no `var`? 😄

Eilon: Can put this on multiple lines too. Not sure where this awful convention came from in this file...

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
```diff
@@ -168,3 +152,42 @@ a {
         color: #13709e;
         text-decoration: underline;
     }
+
+.showRawException {
+    font-size: 1.2em;
+    cursor: pointer;
```
Eilon: Is the cursor style needed?

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
```diff
@@ -168,3 +152,42 @@ a {
...
+    color: #44c5f2;
+    border: none;
+    text-align: left;
+    text-decoration: none;
+    display: inline-block;
+    background-color: white;
```
Eilon: Transparent?

Eilon:
Overall looks **awesome**! I wonder if for the source code we should put some subtle background color? I tried adding `background-color: #ffc;` (same exact color as ASP.NET non-Core uses) to the `#stackpage .source ol li` style and I think it looks a bit nicer. What do you think?

Eilon:
ysod

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
kichalla: Since its a button now (but it doesn't show up as a button in the UI and it appears as a link) the cursor is needed to make it appear as a link.

kichalla:
The background color looks nice. I will send one more iteration with this change.

kichalla:
Updated UI: 
[Internal Server Error.zip](https://github.com/aspnet/Diagnostics/files/389288/Internal.Server.Error.zip)

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
Eilon: Ah ok.

Eilon: Any reason not to keep that one as a link? Links are also automatically tab stops. Might need less CSS that way?

kichalla: Ah ok. For some reason I did not see the tab stop happen and so thought of changing it. Will investigate

pranavkm:
We should start pasting the generated html in jsfiddle. zips are icky

pranavkm:
```
    #stackpage .source ol {
        padding-left: 30px;
    }

.expandCollapseButton {
    cursor: pointer;
    float: left;
    background-size: contain;
    height: 16px;
    width: 16px;
    font-size: 10px;
    padding: 0;
    border: 0;
    margin-top: 8px;
    background-color: #eee;
    position: absolute; 
    left: 10px; 
    margin: 0
}
```

```
<h2 class="stackerror">ArgumentException: Empty path name is not legal.</h2>
                        <ul>
                            <li class="frame" id="frame1">
                                    <h3>MvcSandbox.Controllers.HomeController.RuntimeException(bool innerEx) in <code title="C:\github\mvc\samples\MvcSandbox\Controllers\HomeController.cs">HomeController.cs</code></h3>
                                    <button style="position: absolute; left: 1em; margin: 0" class="expandCollapseButton" data-frameid="frame1">+</button>
                                    
                                    <div class="source">
                                            <ol start="24" class="collapsible" style="display: none;">
```

kichalla:
Thanks @pranavkm I updated the UI now: https://jsfiddle.net/r82ekvoa/

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
kichalla: The issue seems to happen if I use an anchor tag without the `href` attribute. Was reading on the web that tab order is one of the side effects of removing this attribute. Having a value like # for the href would result in the url being modified which i think we might not prefer, so probably button is a better approach here.

kichalla:
:up: :date: https://jsfiddle.net/kazaurt8/

pranavkm:
Could we remove the yellow background? It reminds me of the 90s

Eilon:
Ok fine let's use @pranavkm 's proposed colors...

pranavkm:
#differentshadesofgrey

kichalla:
:up: :date: https://jsfiddle.net/y8eLhnqb/

pranavkm:
:shipit:

Eilon:
:shipit:

PR #325, opened 2016-08-03
[Fixes #324] Diagnostics middleware no longer has a way to expand the…

### LINKED ISSUE #324: Diagnostics middleware no longer has a way to expand the source for compilation errors
There are two separate error pages - one for compilation failures and one for all other exceptions which share the same css and js. In changing the layout for the regular exception page, we've lost the ability to expand \ collapse nodes for the compilation error page. E.g.

image

kichalla:
… source for compilation errors

@pranavkm @Eilon 

https://jsfiddle.net/oLeL8tfo/
https://jsfiddle.net/bv9y73bp/


kichalla:
ping

Eilon:
:shipit:

PR #331, opened 2016-09-06
Use TaskCache class from Microsoft.Extensions.TaskCache.Sources

pakrym:
Instead of Task.FromResult(0)

@BrennanConroy


BrennanConroy:
:shipit:

pranavkm:
Where's the perf?

pakrym:
@pranavkm perf?

PR #332, opened 2016-09-08
Set background color for ErrorPage

andyedinborough:
When using various browser developer tools with a dark theme, the text for the page is unreadable, because the background color has not been set, while the foreground color has.

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.Designer.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/CompilationErrorPage.Designer.cs
@@ -75,6 +75,7 @@ using Microsoft.AspNetCore.Diagnostics.RazorViews
     font-family: 'Segoe UI', Tahoma, Arial, Helvetica, sans-serif;
     font-size: .813em;
     color: #222;
+    background-color: #fff;
 }
 
 h1, h2, h3, h4, h5 {
@@ -127,9 +128,9 @@ body .location {
         cursor: pointer;
     }
 
-    #header .selected {
-        ba");
-            WriteLiteral(@"ckground: #44c5f2;
+    #h");
+            WriteLiteral(@"eader .selected {
+        background: #44c5f2;
         color: #fff;
     }
 
@@ -186,9 +187,9 @@ body .location {
 
 .page table {
     border-collapse: separate;
-    border-spacing: 0;
-    margin:");
-            WriteLiteral(@" 0 0 20px;
+    bo");
+            WriteLiteral(@"rder-spacing: 0;
+    margin: 0 0 20px;
 }
 
 .page th {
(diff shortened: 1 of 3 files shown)


andyedinborough:
For example: 
image

pranavkm:
You have to regenerate the designer files for this to take effect. The css files get inlined in to the designer file as part of being generated.

pranavkm:
cc @kichalla - could you add instructions on running RazorPageGenerator?

kichalla:
When you are under the project folder, you can use the `RazorPageGenerator` tool like below:

```
Usage:   dotnet razorpagegenerator <root-namespace-of-views>
Example: dotnet razorpagegenerator Microsoft.AspNetCore.Diagnostics.RazorViews
```

andyedinborough:
Done.  Let me know if I did that right ;]

pranavkm:
Might be the result of using a newer version of Razor but the designer formats some of the other generated artifacts ever so slightly:

image

Would you mind running `samples/DeveloperExceptionPageSample` and pasting in the generated error page into jsfiddle or liveweave and paste it in here? Just want to make sure the page's rendering correctly.

andyedinborough:
Done:  https://jsfiddle.net/andyedinborough/fe15nty2/
image

pranavkm:
This looks pretty great. I was talking to @sebastienros offline and he recommended doing more and using a css reset script (something like http://html5doctor.com/html-5-reset-stylesheet) so similar themeing issues don't affect the diagnostics page. I'm not super sure we need all of that. Let me run it by @Eilon for his thoughts on this.

Eilon:
Eh, I don't think we need the full reset script because this is only needed for a more obscure dev scenario. I'm fine just setting the white background, though.

Don't we need to update this in all our error pages, though? @kichalla ?

andyedinborough:
If you'd like to see it with a reset, I created https://github.com/andyedinborough/Diagnostics/tree/add-normalize, https://jsfiddle.net/andyedinborough/3y3defht/.

image

sebastienros:
If you decide to set the background, then you have to also set the foreground color. Then by extension you should set the size, the line-height, the weight, and probably the font (with fall-backs). For each of the styles the page uses: headers, code, ...

Eilon:
@sebastienros I don't entirely agree with that, though I agree with your statement in general. The only supported scenario for this page is to have it "full screen" in a browser window. Browsers don't need reset scripts. It just so happens that in some obscure (i.e. not-really-supported scenarios) there is a need to reset some styles.

andyedinborough:
The foreground has been set, and that's the problem.  The font-size and weight have been set also.

sebastienros:
> in some obscure (i.e. not-really-supported scenarios) there is a need to reset some styles

I agree, and when resetting "some styles" you need to reset the related ones. For instance background goes in pair with foreground. And this was the actual issue, the foreground was set and not the background. So I agree with the solution as the two of then are set.

I would still suggest to set the other elements I mentioned, to have a deterministic render.

> Browsers don't need reset scripts

We'll talk about this one offline ;)

pranavkm:
Merging in as is. Thanks for the PR!

PR #333, opened 2016-09-12
fix incorrect pin in 1.0.0 tag

anurse:
The `1.0.0` tag has a version of `1.0.0-rtm-21431` for `PageGenerator`, but we actually "shipped" (to our internal feeds) `1.0.0`. This updates the pin to use the correct version so that future builds can accurately locate the previously "released" package on our internal feeds.

/cc @Eilon @pranavkm 

When this is merged, I'll update the `1.0.0` tag to point to the new commit.

### DIFF
--- a/src/PageGenerator/project.json
+++ b/src/PageGenerator/project.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-rtm-21431",
+  "version": "1.0.0",
   "description": "Builds the pages for the Diagnostics projects. Runs in build.cmd.",
   "buildOptions": {
     "warningsAsErrors": true,


anurse:
Travis failure is because this branch doesn't have the OpenSSL on Mac/Homebrew fix (as with all the other 1.0.0 tags...)

pranavkm:
:shipit:

pranavkm:
Were you trying to regen the pages? I thought the tool typically doesn't get in the regular build pipeline.

anurse:
No, but because this is in the `src` directory, the train build tools think we need to fetch it from the releases feed when we're not building this repo, in order to ensure tests will have everything them need. (even though you are correct, that it is not needed)

Issue #292, opened 2016-05-13
Finding Original RequestUri with IExceptionHandlerFeature


Tratcher:
_From @shawnwildermuth on May 12, 2016 19:42_

I'm trying to log the request URI when an exception is thrown. Logging is too abstracted out to do it so I thought that IHttpRequestFeature was my trick. Here's what I'm trying, please let me know if I'm way off the right track:

First I'm using an exception page:

```
app.UseExceptionPage("/Exception");
```

This logs the exception, but just the bare exception, no state of the request.

So, inside the Action for /Exception, I'm doing this:

```
    [HttpGet("Exception")]
    public IActionResult Exception()
    {
      var exception = HttpContext.Features.Get<IExceptionHandlerFeature>();
      var request = HttpContext.Features.Get<IHttpRequestFeature>();

      if (exception != null && request != null)
      {
        var originalPath = request.Path;
        // I could cast the request object to a Frame to get at the information, but that feels dirty

        // Manually logging the information
      }

      return View();
    }
```

_Copied from original issue: aspnet/HttpAbstractions#630_

Tratcher:
We should add the original path to IExceptionHandlerFeature here:
https://github.com/aspnet/Diagnostics/blob/0a444088c9a7c5c6b4073c92104b48af734ef523/src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L66

bedwar2:
When a 403 is encountered I'd like to get the RequestURI to log what URL user attempted.  Shouldn't ReuestURI be on the IHTTPRequestFeature in that case as well?  I see the original request was for it to be on IHttpRequestFeature.  I still think it should be there.  I use app.UseStatusCodePagesWithReExecute.

Tratcher:
@Eilon are the versioning guildlines ready for v1.1? Do they allow adding properties to feature interfaces?

Eilon:
@Tratcher working on the guidelines. Changing an interface in any way is a breaking change that we would not make. We can add a new interface, though.

PR #334, opened 2016-09-12
Added Path property to IExceptionHandlerFeature 

mikaelm12:
In response to #292
The argument for not implementing a second interface and changing the contract on IExceptionHandlerFeature as I understand it is that essentially no one should run into this problem and that most likely only the ExceptionHandlerMiddleware is implementing the interface.
@Tratcher @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
@@ -9,6 +9,6 @@ namespace Microsoft.AspNetCore.Diagnostics
     {
         Exception Error { get; }
 
-        String Path { get; }
+        string Path { get; }
     }
 }
--- a/src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
@@ -8,6 +8,7 @@ namespace Microsoft.AspNetCore.Diagnostics
     public interface IExceptionHandlerFeature
     {
         Exception Error { get; }
+
         String Path { get; }
     }
 }
--- a/src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
@@ -8,5 +8,6 @@ namespace Microsoft.AspNetCore.Diagnostics
     public interface IExceptionHandlerFeature
     {
         Exception Error { get; }
+        String Path { get; }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
```diff
@@ -8,5 +8,7 @@ namespace Microsoft.AspNetCore.Diagnostics
     public interface IExceptionHandlerFeature
     {
         Exception Error { get; }
+
+        string Path { get; }
```
pranavkm: We can't change the interface for `1.1`.  This would be a breaking change.

Tratcher: Yes it's breaking, but is it an acceptable break? There's no expectation that anyone implement this interface, and even a github search only shows consumers.

pranavkm: > but is it an acceptable break

I have no idea. @Eilon usually doesn't like any sort of breaking changes in minor releases.

An alternate way to fix this without breaking would be to introduce a new interface and add the property on there.

Eilon: Not an acceptable break. Add a new interface.

src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs
```diff
@@ -65,6 +65,7 @@ public class ExceptionHandlerMiddleware
                     var exceptionHandlerFeature = new ExceptionHandlerFeature()
                     {
                         Error = ex,
+                        Path = originalPath.Value
```
Eilon: Can add a trailing comma here... though given the other comments in this PR, this code might change anyway.

src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerFeature.cs
Tratcher: IExceptionHandlerPathFeature?

Eilon: Yeah sure.

src/Microsoft.AspNetCore.Diagnostics.Abstractions/IExceptionHandlerPathFeature.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Diagnostics
+{
+    public interface IExceptionHandlerPathFeature : IExceptionHandlerFeature
+    {
+        string Path { get; }
```
Eilon: Nice, add doc comments to the property and the type.

pranavkm: Would it make sense to make this a Context type? Just so that we don't have to rev the interface any time we want to change it?

Tratcher: No, the Features are all interfaces so they can be implemented by anybody.

Tratcher: Copy the doc comment from IHttpReqeuestFeature.Path

PR #341, opened 2017-01-06
Remove Microsoft.AspNetCore.DiagnosticsViewPage.Sources package

natemcmaster:
As far as I can tell, this package is not used outside of this repo. This moves shared code into a folder instead of using a 'shared sources project'.

This is part of the preparation for VS 2017 upgrade. 

This is the last of the "shared" sources packages. Resolves https://github.com/aspnet/Coherence-Signed/issues/379

Related change: https://github.com/aspnet/Coherence/commit/b656b15e03b65247d4ee072717c80c37c50799bd


natemcmaster:
Ping @pranavkm, you okay with this change?

PR #352, opened 2017-03-09
[Fixes #344] DeveloperExceptionPage throws when SourceFileContent is null in an ICompilationException

### LINKED ISSUE #344: DeveloperExceptionPage throws when SourceFileContent is null in an ICompilationException
I hit this case while working through some Razor issues. 

The SourceFileContent is dereferenced here: https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs#L138

We should probably harden this code since it's an error page.

kichalla:


src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -126,42 +126,66 @@ private Task DisplayException(HttpContext context, Exception ex)
...
+            {
+                Model = model
+            };
+
+            if (compilationException.CompilationFailures == null)
+            {
+                return errorPage.ExecuteAsync(context);
+            }
+
             foreach (var compilationFailure in compilationException.CompilationFailures)
```
pranavkm: Could we refactor this in a different method so we don't have two different calls to `ExexcuteAsync`. ie

```
var errorPage = new CompilationErrorPage
{
   Model = model
};

ReadCompilationErrors(model);
return errorPage.ExecuteAsync(context);
```

test/Microsoft.AspNetCore.Diagnostics.Tests/DeveloperExceptionPageMiddlewareTest.cs
```diff
@@ -49,5 +51,105 @@ public class DeveloperExceptionPageMiddlewareTest
...
+                });
+            var server = new TestServer(builder);
+            var listener = new TestDiagnosticListener();
+            diagnosticListener.SubscribeWithAdapter(listener);
+
+            // Act
+            await server.CreateClient().GetAsync("/path");
+
+            // This ensures that all diagnostics are completely written to the diagnostic listener
+            Thread.Sleep(1000);
```
pranavkm: `await Task.Delay(1000);`

davidfowl: Dude, is there no way to avoid a sleep in tests?

PR #358, opened 2017-04-16
Disable SqlServer specific tests on netcoreapp2.0 on win7/2008

smitpatel:
@mikeharder @ryanbrandenburg 
This disables tests which are facing same issue as EF tests.
Manually verified on win7 VM.


smitpatel:
Follow up issue #359

PR #360, opened 2017-04-17
Deleted obsolete razor views and models

### LINKED ISSUE #295: Disambiguate duplicate Microsoft.AspNetCore.DiagnosticsViewPage.Views.* type names
The types in this namespace come from https://github.com/aspnet/Diagnostics/tree/dev/src/Microsoft.AspNetCore.DiagnosticsViewPage.Sources and are included in two other projects. This means we have the exact same public type names in two different assemblies, which isn't ideal. Nothing actual bad happens, but it's a bad practice.

We should look into one of these:
1. If possible, make the type internal. Case closed.
2. If they must be public, put the types in different namespaces based on where the types are included. E.g. use an `#ifdef` in the `.Sources` package to do something like `#if EF 'namespace foo.EF' #elseif ASPNET 'namespace foo.AspNet' #else #error #endif`

**Update1 (kichalla):**
New path: https://github.com/aspnet/Diagnostics/tree/dev/src/shared

kichalla:
[Fixes #295 ] Disambiguate duplicate Microsoft.AspNetCore.DiagnosticsViewPage.Views.* type names #295


kichalla:
ping

Issue #366, opened 2017-04-25
Flaky Test: EntityFrameworkCore.Tests.DatabaseErrorPageMiddlewareTest.Pass_thru_when_context_not_in_services


mikeharder:
```
[Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.Tests.DatabaseErrorPageMiddlewareTest.Pass_thru_when_context_not_in_services] Assert.Throws() Failure
Expected: typeof(System.Data.SqlClient.SqlException)
Actual:   typeof(System.InvalidOperationException): An exception has been raised that is likely due to a transient failure. If you are connecting to a SQL Azure database consider using SqlAzureExecutionStrategy.
   at Microsoft.EntityFrameworkCore.Storage.Internal.SqlServerExecutionStrategy.Execute[TState,TResult](Func`2 operation, Func`2 verifySucceeded, TState state)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(IReadOnlyList`1 entriesToSave)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.Tests.DatabaseErrorPageMiddlewareTest.ContextNotRegisteredInServicesMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.MigrationsEndPointMiddleware.<Invoke>d__5.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.<Invoke>d__6.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.<Invoke>d__6.MoveNext()
[...]

PR #369, opened 2017-04-26
Treat all SQL exceptions the same in DatabaseErrorPageMiddlewareTest

AndriySvyryd:


test/Diagnostics.EFCore.FunctionalTests/DatabaseErrorPageMiddlewareTest.cs
```diff
@@ -322,7 +322,20 @@ public ApplyMigrationsMiddleware(RequestDelegate next)
             var server = new TestServer(builder);
 
             var ex = await Assert.ThrowsAsync<SqlException>(async () =>
-                await server.CreateClient().GetAsync("http://localhost/"));
+            {
+                try
+                {
+                    await server.CreateClient().GetAsync("http://localhost/");
+                }
+                catch (InvalidOperationException ope)
```
pranavkm:
```
catch (InvalidOperationException exception) when (exception.InnerException != null)
{
  throw exception.InnerException;
}
```

pranavkm:
@Eilon should this go in preview1? Test stabilization and all that.

Eilon:
Have we seen this test fail recently? If so, I'm fine porting this to preview1. @ajcvickers thoughts?

Eilon:
Ah just saw the related bug https://github.com/aspnet/Diagnostics/issues/366#issuecomment-297567879

Definitely looks worth porting to me.

ajcvickers:
@Eilon @AndriySvyryd Looks like we should port it.

AndriySvyryd:
Done

PR #370, opened 2017-04-27
React to Logging API changes

BrennanConroy:


BrennanConroy:
Ping

src/Microsoft.AspNetCore.Diagnostics.Elm/ElmExtensions.cs
```diff
@@ -22,9 +21,12 @@ public static IApplicationBuilder UseElmCapture(this IApplicationBuilder app)
             }
 
             // add the elm provider to the factory here so the logger can start capturing logs immediately
-            var factory = app.ApplicationServices.GetRequiredService<ILoggerFactory>();
-            var provider = app.ApplicationServices.GetRequiredService<ElmLoggerProvider>();
-            factory.AddProvider(provider);
+            var factory = app.ApplicationServices.GetRequiredService<ILoggerFactory>() as LoggerFactory;
```
pranavkm: Why is this check necessary?

samples/MiddlewareAnalysisSample/Startup.cs
```diff
@@ -85,6 +85,11 @@ public void Configure(IApplicationBuilder app, ILoggerFactory factory, Diagnosti
         public static void Main(string[] args)
         {
             var host = new WebHostBuilder()
+                .ConfigureLogging((_, factory) =>
```
pranavkm: Is this for tests?

src/Microsoft.AspNetCore.Diagnostics.Elm/ElmExtensions.cs
BrennanConroy: Because `ILoggerFactory` is obsoleting `AddProvider` it's now just a method on `LoggerFactory`

samples/MiddlewareAnalysisSample/Startup.cs
BrennanConroy: This is to configure `LoggerFactory` this overload will be called with the concrete `LoggerFactory`

PR #371, opened 2017-04-28
Missed some obsolete logging calls

BrennanConroy:


BrennanConroy:
@Eilon

BrennanConroy:
Should we file an issue to look at wrapping the factory?

Eilon:
@BrennanConroy - yes I think we should. Can you do that?

PR #380, opened 2017-05-23
Target netstandard2.0

pranavkm:


src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
```diff
@@ -69,6 +69,16 @@ public DatabaseErrorPageMiddleware([NotNull] RequestDelegate next, [NotNull] ISe
 
             try
             {
+#if NET461
+                // TODO This probably isn't the correct place for this workaround, it
```
pranavkm: @bricelam - what's up with this?

src/Microsoft.AspNetCore.Diagnostics.Elm/baseline.netframework.json
```diff
@@ -0,0 +1,2210 @@
+{
```
natemcmaster: IIRC correctly API check looks for `baseline.netcore.json` when the project targets .NET Standard, so this file needs to be renamed.

src/Microsoft.AspNetCore.MiddlewareAnalysis/baseline.netframework.json
```diff
@@ -0,0 +1,218 @@
+{
```
natemcmaster: ditto

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
pranavkm: Also cc @ajcvickers

ajcvickers: @bricelam Do you know anything about this? Can it be removed?

bricelam: No idea. I vaguely remember something about the response's content type not being set correctly without this.

It was written ages ago and has probably since been fixed. I say we remove it and see what breaks.

PR #396, opened 2017-07-18
Remove NetStandardLibraryImplicitPackageVersion

mikeharder:
- For consistency with other repos (e.g. https://github.com/aspnet/Razor/blob/rel/2.0.0/build/common.props#L22)

I created this PR against 2.0 since it's nice to have consistent `*.props` files in case they need to be updated for servicing.  However, I'm also fine merging this to dev but not 2.0.


mikeharder:
@Eilon: Take this for 2.0 or dev?

PR #402, opened 2017-07-29
update get full exception info when exception has InnerException

fs7744:
when exception has InnerException
use exception.Message will not show full info
sometimes the InnerException will be very helpfull

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.Diagnostics
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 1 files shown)


fs7744:
The MessageFormatter in [LoggerExtensions](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs) only keep the second parameter, never use the parameter ex , so the InnerException will be ignored

Tratcher:
Then the issue should be addressed in Logging, not here.

fs7744:
I create a issue in https://github.com/aspnet/Logging/issues/675

fs7744:
Updated.
Here's some background #442. It's an unfortunate design flaw of logging. 
So all logger must be handle exception and user's message both.
Although the code will log redundant ex.Message in https://github.com/fs7744/Diagnostics/blob/62cc1bd80bc9eeaf78189f68936f58e5e98c5239/src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L48
But it not big problem.

PR #408, opened 2017-10-10
Add abstractions for Health Checks and a simple middleware and service to run them

analogrelay:
This PR adds a simple abstraction around health checks (`IHealthCheck`) which is designed to be used throughout ASP.NET. There is also a simple service to collect Health Checks out of DI and make them runnable, and a middleware to allow exposing that information via a JSON endpoint.

Other health check systems can contribute by:

1. Adding health checks via implementing `IHealthCheck`
1. Replacing the `IHealthCheckService` with something with more functionality
1. Replacing the middleware with one that supports more functionality

In each case, replacing one part should still allow them to participate in the other parts. For example, replacing the middleware alone would still allow the system to use the existing health checks registered by framework components.

We also plan to add health checks within the framework, so having a light abstraction for that purpose is important, and would allow other health check systems to plug in to framework-provided health checks.

Based on the code in https://github.com/dotnet-architecture/HealthChecks, but with some reduced functionality to simplify logic.

CC @muratg @Eilon


src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -0,0 +1,16 @@
+using Microsoft.AspNetCore.Diagnostics.HealthChecks;
```
muratg: Nit: License notice.  (/Troy)

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,75 @@
+using System.Linq;
```
muratg: Ditto (and other files too)

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
anurse: Fixed

samples/HealthChecksSample/Program.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            BuildWebHost(args).Run();
+        }
+
+        public static IWebHost BuildWebHost(string[] args) =>
+            new WebHostBuilder()
```
glennc: Reason to not use `CreateDefaultBuilder`?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Diagnostics.HealthChecks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class HealthCheckAppBuilderExtensions
+    {
+        public static IApplicationBuilder UseHealthChecks(this IApplicationBuilder applicationBuilder, PathString path)
```
glennc: Overload that accepts options func?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            // TODO: Configurable
+            if (context.Request.Path == _healthCheckOptions.Path)
+            {
+                // TODO: Caching
+                // Get results
+                var result = await _healthCheckService.CheckHealthAsync(context.RequestAborted);
+
+                // Map status to response code
+                switch (result.Status)
```
Drawaes: https://httpstatuses.com/429

That is used by consul.io and others to say "backoff I am overloaded"

anurse:
The usage pattern is the following:

```csharp
// In some ConfigureServices:
services.AddHealthChecks()
  .AddCheck("Foo", () => ...)
  .AddSomeLibraryProvidedCheck("Bar")
// ... etc ...
// Later on or in some other ConfigureServices
services.AddHealthChecks()
  .AddCheck("Baz", () => ...)
```

The `.AddHealthChecks` call just adds the `IHealthCheckService` instance to the container and then returns a builder to register individual checks (checks are just `IHealthCheck` instances in DI). `AddHealthChecks` is idempotent, so you can call it in multiple places and it only adds the `IHealthCheckService` once, but you can use it to add additional checks (like Authentication)

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: I'm not sure we want to specifically use that error code, as some systems use https://httpstatuses.com/503. Health checks may not be directly related to the number of requests. Regardless: we do want to make this output somewhat configurable and even if we don't use 429 by default, we'll want to make it an option. Perhaps a dictionary mapping statuses to HTTP status codes, or some delegate handler thing.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    case HealthCheckStatus.Failed:
```
glennc: Many things doing the probing only know about 200 or not 200. But that doesn't mean we shouldn't provide more fine grained if it makes sense to. 

Failed is weird, it means something bad happened in health check implementations and it couldn't be handled or something. Right?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+                    case HealthCheckStatus.Healthy:
+                        context.Response.StatusCode = StatusCodes.Status200OK;
+                        break;
+                    default:
+                        break;
+                }
+
+                // Render results to JSON
+                // TODO: Authentication policy to control this? Sensitive data hiding? etc.
+                var json = new JObject(
```
glennc: This needs to be able to be turned on and off.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: This being a first pass, I'll file issues for some of these issues after the PR is merged. https://github.com/aspnet/Diagnostics/issues/409

Drawaes: 429 is standard for say consul.io which is pretty massive. There is 200 429 and failure  (everything else and just not there). So it would be key... but sure some mapping would be fine.

anurse:
> Need a way to say "don't dump out the keys unless its anything but good", As discussed changing healthchecks are a pain as they trigger "updates" in things like consul.

Yep, I should file the issue I planned to file after this PR for exactly this :). https://github.com/aspnet/Diagnostics/issues/410

src/Microsoft.Extensions.Diagnostics.HealthChecks/CompositeHealthCheckResult.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.Extensions.Diagnostics.HealthChecks
+{
+    /// <summary>
+    /// Represents the results of multiple health checks
+    /// </summary>
+    public class CompositeHealthCheckResult
```
glennc: Is there an interesting ToString we can do here?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
glennc: This will be a constant tension of simplicity vs configurableness. You could envision an options type that contains a map of HealthCheckStatus to status code that you could change. Assuming it didn't look terrible to use.

samples/HealthChecksSample/Program.cs
anurse: Because I didn't. Can change.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
anurse: No options right now. But as we add some, sure.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: It means that a health check threw an unhandled exception, basically.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: https://github.com/aspnet/Diagnostics/issues/410

src/Microsoft.Extensions.Diagnostics.HealthChecks/CompositeHealthCheckResult.cs
anurse: I'm hesitant to put JSON in here, but maybe we could do something.

Tratcher:
What kind of health checks did you have in mind for building into ASP.NET?

samples/HealthChecksSample/Program.cs
Tratcher: You can't, none of our repos can use CreateDefaultBuilder without creating a circular dependency.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
Tratcher: Aren't Options all supposed to be configured as services now? Most Use methods no longer take parameters.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Microsoft.AspNetCore.Diagnostics.HealthChecks
+{
+    public class HealthCheckMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly HealthCheckOptions _healthCheckOptions;
+        private readonly IHealthCheckService _healthCheckService;
+
+        public HealthCheckMiddleware(RequestDelegate next, HealthCheckOptions healthCheckOptions, IHealthCheckService healthCheckService)
```
Tratcher: IOptions

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        /// <summary>
+        /// Process an individual request.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public async Task InvokeAsync(HttpContext context)
+        {
+            // TODO: Configurable
```
Tratcher: Isn't it already configurable via options?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/Microsoft.AspNetCore.Diagnostics.HealthChecks.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <Description>ASP.NET Core middleware for returning the results of Health Checks in the application</Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>diagnostics;healthchecks</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" />
```
Tratcher: Hosting.Abstractions rather than Http.Abstractions?

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckStatus.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <remarks>
+    /// The values of this enum or ordered from least healthy to most healthy. So <see cref="HealthCheckStatus.Degraded"/> is
+    /// greater than <see cref="HealthCheckStatus.Unhealthy"/> but less than <see cref="HealthCheckStatus.Healthy"/>.
+    /// </remarks>
+    public enum HealthCheckStatus
+    {
+        /// <summary>
+        /// This value should not be returned by a health check. It is used to represent an uninitialized value.
+        /// </summary>
+        Unknown = 0,
```
Tratcher: `None` is the normal enum default.

anurse:
@Tratcher In theory, EF could put one in to check the database whenever a DbContext is available. Ones could exist for monitoring metrics (concurrent requests counts, timeouts, request duration, etc.) and returning unhealthy results when they go above certain thresholds as well. Spring has a model where any component you plug in can just contribute Health Indicators (as they call them) so we want to provide a very light abstraction (see the size of HealthChecks.Abstractions and the fact that it doesn't have `IHealthCheckService`) so that any library will feel confident providing Health Checks.

All of this is theoretical right now, but we have some concrete examples from other platforms to build on.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: My thinking would be a `Func<CompositeHealthCheckResult, int>` or something. Similarly, a `Func<HttpContext, CompositeHealthCheckResult, bool>` to determine if details should be rendered (need `HttpContext` to determine if the user is authorized, if that's how you want to do it). Then we could also provide extension methods on `HealthCheckMiddlewareOptions` to tie in with other systems (`.ShowDetailsOnlyWhenUnhealthy()` or `.ShowDetailsToAuthorizedUsersOnly(policy: "Foo")`)

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckStatus.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// This value should not be returned by a health check. It is used to indicate that an unexpected exception was
+        /// thrown when running the health check.
+        /// </summary>
+        Failed = 1,
+
+        /// <summary>
+        /// Indicates that the health check determined that the component was unhealthy
```
Eilon: ⬇️ Since we already have doc comments here, might as well end these sentences with periods.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+                // Render results to JSON
+                // TODO: Authentication policy to control this? Sensitive data hiding? etc.
+                var json = new JObject(
+                    new JProperty("status", result.Status.ToString()),
+                    new JProperty("results", new JObject(result.Results.Select(pair =>
+                        new JProperty(pair.Key, new JObject(
+                            new JProperty("status", pair.Value.Status.ToString()),
+                            new JProperty("description", pair.Value.Description),
+                            new JProperty("data", new JObject(pair.Value.Data.Select(p => new JProperty(p.Key, p.Value))))))))));
+                await context.Response.WriteAsync(json.ToString(Formatting.None));
```
Eilon: What's the security model for health checks? What prevents a random person from getting the detailed health status of a server?

Eilon (changes requested):

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: See #410

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckStatus.cs
anurse: Periods are for the weak you should just be able to figure out where my sentences end without me having to tell you :P

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// The key maps to the <see cref="IHealthCheck.Name"/> property of the health check, and the value is the <see cref="IHealthCheck"/>
+        /// instance itself.
+        /// </remarks>
+        public IReadOnlyDictionary<string, IHealthCheck> Checks { get; }
+
+        /// <summary>
+        /// Constructs a <see cref="HealthCheckService"/> from the provided collection of <see cref="IHealthCheck"/> instances.
+        /// </summary>
+        /// <param name="healthChecks">The <see cref="IHealthCheck"/> instances that have been registered in the application.</param>
+        public HealthCheckService(IEnumerable<IHealthCheck> healthChecks) : this(healthChecks, NullLogger<HealthCheckService>.Instance) { }
```
Tratcher: Why is the logger optional?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        /// <summary>
+        /// Constructs a <see cref="HealthCheckService"/> from the provided collection of <see cref="IHealthCheck"/> instances, and the provided logger.
+        /// </summary>
+        /// <param name="healthChecks">The <see cref="IHealthCheck"/> instances that have been registered in the application.</param>
+        /// <param name="logger">A <see cref="ILogger{T}"/> that can be used to log events that occur during health check operations.</param>
+        public HealthCheckService(IEnumerable<IHealthCheck> healthChecks, ILogger<HealthCheckService> logger)
+        {
+            // Scan the list for duplicate names to provide a better error if there are duplicates.
+            var names = new HashSet<string>();
```
Tratcher: Specify comparer.

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckStatus.cs
Eilon: I don't understand

What are you? talking about

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }
+
+            if(duplicates.Count > 0)
+            {
+                throw new InvalidOperationException($"Duplicate health checks were registered with the name(s): {string.Join(", ", duplicates)}");
+            }
+
+            Checks = healthChecks.ToDictionary(c => c.Name);
+            _logger = logger;
```
Tratcher: Null check

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        /// <summary>
+        /// Runs all the health checks in the application and returns the aggregated status.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> which can be used to cancel the health checks</param>
+        /// <returns>
+        /// A <see cref="Task{T}"/> which will complete when all the health checks have been run,
+        /// yielding a <see cref="CompositeHealthCheckResult"/> containing the results.
+        /// </returns>
+        public async Task<CompositeHealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default)
```
Tratcher: Do we want to implement any concurrency throttling? E.g. having the service only allow one pass at a time via async lock. This may be related to your caching TODO. @glennc has already shared stories about DOS via heath checks.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            var results = new Dictionary<string, HealthCheckResult>(Checks.Count);
+            foreach (var pair in Checks)
+            {
+                // If the health check does things like make Database queries using EF or backend HTTP calls,
```
Tratcher: ThrowIfCancellationRequested to abort the loop?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,118 @@
...
+                // it may be valuable to know that logs it generates are part of a health check. So we start a scope.
+                using (_logger.BeginScope(new HealthCheckLogScope(pair.Key)))
+                {
+                    HealthCheckResult result;
+                    try
+                    {
+                        _logger.LogTrace("Running health check: {healthCheckName}", pair.Key);
+                        result = await pair.Value.CheckStatusAsync(cancellationToken);
```
Tratcher: Should there be a configurable timeout per check? Though I guess you could do that on the options for that check.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,118 @@
...
+                        _logger.LogTrace("Health check '{healthCheckName}' completed with status '{healthCheckStatus}'", pair.Key, result.Status);
+                    }
+                    catch (Exception ex)
+                    {
+                        // We don't log this as an error because a health check failing shouldn't bring down the active task.
+                        _logger.LogTrace(ex, "Health check '{healthCheckName}' threw an unexpected exception", pair.Key);
```
Tratcher: Debug?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthChecksBuilderAddCheckExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        /// <summary>
+        /// Adds a new health check with the specified name and implementation.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHealthChecksBuilder"/> to add the check to.</param>
+        /// <param name="name">The name of the health check, which should indicate the component being checked.</param>
+        /// <param name="check">A delegate which provides the code to execute when the health check is run.</param>
+        /// <returns>The <see cref="IHealthChecksBuilder"/></returns>
+        public static IHealthChecksBuilder AddCheck(this IHealthChecksBuilder builder, string name, Func<Task<HealthCheckResult>> check) =>
```
Tratcher: What about `AddCheck<T>() where T is IHealthCheck` and `AddCheck<T, O>(Action<O> configureOptions) where T is IHealthCheck` and O is the Options for T.

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
```diff
@@ -0,0 +1,34 @@
...
+    public interface IHealthCheckService
+    {
+        /// <summary>
+        /// A <see cref="IReadOnlyDictionary{TKey, T}"/> containing all the health checks registered in the application.
+        /// </summary>
+        /// <remarks>
+        /// The key maps to the <see cref="IHealthCheck.Name"/> property of the health check, and the value is the <see cref="IHealthCheck"/>
+        /// instance itself.
+        /// </remarks>
+        IReadOnlyDictionary<string, IHealthCheck> Checks { get; }
```
Tratcher: Why does this need to be exposed publicly? Only the service itself consumes them, no?

Tratcher: Has there been any discussion on the ability to add and remove checks at runtime?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
Eilon: Loggers are not normally required, are they? It's conceptually fine for them to be missing...

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckStatus.cs
Eilon: It depends on the circumstance. I think `None` doesn't make much sense here...

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
Drawaes: Trace is good. Debug is a pain. Again you are debugging your app and it gets 3 or 4 lines every second due to health checks, it's just annoying.  Although in saying that a lot of places health checks are actually used filtering is heavily used.

I love .net 2.0 filtering FYI ;)

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
anurse: Callers may want to run individual health checks or groups of health checks. The previous system had methods and infrastructure for this, but we didn't want to add that complexity in the full product. However, we still want consumers to be able to access and run individual checks.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
anurse: Especially here when the HealthCheckService will eventually be used by an ASP.NET 4.x component as well.

Drawaes:
Punctuation is for people with bad parsing logic ;)

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: Do our middlewares use IOptions? It looks like some in aspnet/BasicMiddleware do, but others such as Proxy and Authentication don't. Middleware don't come from DI, but I suppose they are activated from it. I can use IOptions.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: Yep, out of date comment.

src/Microsoft.Extensions.Diagnostics.HealthChecks/CompositeHealthCheckResult.cs
Eilon: Yeah be careful about making it look like ToString returns some kind of legitimate info as if it's a contract that people can rely on. That's not necessarily bad, but contracts have cost, so they have to be "worth it." Else, just put some nice human-readable stuff in it. Or nothing.

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
anurse: > Has there been any discussion on the ability to add and remove checks at runtime?

No, and I don't think that's a feature we've seen in any other systems. It's not really the kind of thing you'd want to add after the app starts.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
anurse: Sure, that does actually fit my general policy of Debug being useful for a human and Trace being only really useful by log analysis tools. :)

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
anurse: Also more complicated than we want to do right now. That's what the sample had, and it got messy and started interfering with everything. Checks don't cache ever in this model, only the middleware caches the check. It could have per-check customization for caching, but that's rarely something you want to do. None of the other systems that I've looked at really do any caching beyond a simple TTL for the full results.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
anurse: I like the idea of preventing simultaneous checks. I'll look in to that.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
Tratcher: Most of our components do require loggers, they don't provide two constructors.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
anurse: If a health check wants to do timeout, they can add that functionality. Adding it to the outer system will add fairly significant complication.

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
Tratcher: We get quite a few requests to be able to add things without app restarts. E.g. middleware, auth handlers, etc..

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
Tratcher: Yes, IOptions is pervasive now.

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
anurse: We'll wait for those requests then :)

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
anurse: The logger is intentionally optional here because of the desire to have a usage pattern that isn't dependent upon DI or our Logging components.

src/Microsoft.Extensions.Diagnostics.HealthChecks/CompositeHealthCheckResult.cs
```diff
@@ -0,0 +1,66 @@
...
+        public CompositeHealthCheckResult(IReadOnlyDictionary<string, HealthCheckResult> results)
+        {
+            Results = results;
+            Status = CalculateAggregateStatus(results.Values);
+        }
+
+        private HealthCheckStatus CalculateAggregateStatus(IEnumerable<HealthCheckResult> results)
+        {
+            // This is basically a Min() check, but we know the possible range, so we don't need to walk the whole list
+            var currentValue = HealthCheckStatus.Healthy;
```
pranavkm: Should this start off as `Unknown` (in the event no results are configured)?

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/IHealthCheck.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Diagnostics.HealthChecks
+{
+    /// <summary>
+    /// Represents a health check, which can be used to check the status of a component in the application, such as a backend service, database or some internal
+    /// state.
+    /// </summary>
+    public interface IHealthCheck
```
pranavkm: `IHealthCheckSource` or something like that since it's the thing that does a check?

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/IHealthCheck.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// Gets the name of the health check, which should indicate the component being checked.
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Runs the health check, returning the status of the component being checked.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the health check.</param>
+        /// <returns>A <see cref="Task{HealthCheckResult}"/> that completes when the health check has finished, yielding the status of the component being checked.</returns>
+        Task<HealthCheckResult> CheckStatusAsync(CancellationToken cancellationToken = default);
```
pranavkm: `CheckHealthAsync` \ `GetResultAsync`? `Status` doesn't appear elsewhere in the API.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            healthChecks = healthChecks ?? throw new ArgumentNullException(nameof(logger));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            // Scan the list for duplicate names to provide a better error if there are duplicates.
+            var names = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var duplicates = new List<string>();
+            foreach (var check in healthChecks)
+            {
+                if(!names.Add(check.Name))
```
pranavkm: formatting

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,121 @@
...
+                {
+                    duplicates.Add(check.Name);
+                }
+            }
+
+            if(duplicates.Count > 0)
+            {
+                throw new InvalidOperationException($"Duplicate health checks were registered with the name(s): {string.Join(", ", duplicates)}");
```
pranavkm: `ArgumentException`

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckResult.cs
```diff
@@ -0,0 +1,188 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Diagnostics.HealthChecks
+{
+    /// <summary>
+    /// Represents the result of a health check.
+    /// </summary>
+    public struct HealthCheckResult
```
pranavkm: Should this expose the source name?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthChecksBuilderAddCheckExtensions.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+// This type is used like this:
```
pranavkm: We do this everywhere, don't think you need these comments.

src/Microsoft.Extensions.Diagnostics.HealthChecks/CompositeHealthCheckResult.cs
anurse: If no checks are configured, the result should be `Healthy`

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/IHealthCheck.cs
anurse: If it's the thing that does the check, is it not a "Health Check" ? :)

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckResult.cs
anurse: Yeah, I thought about that. The somewhat weird thing is that this object is produced by the `IHealthCheck` itself so it seems awkward to have it able to set the name of the result (potentially to something different from it's actual name). The consumer of `CheckHealthAsync` always knows which result maps to which check (because either they called the `IHealthCheck` manually and have it's `Name` available, or via `CompositeHealthCheckResult`'s dictionary) so I don't think it's necessary to have the name here.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthChecksBuilderAddCheckExtensions.cs
anurse: Well, but we don't normally put these in `.DependencyInjection`, see https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging/LoggingBuilderExtensions.cs and ... well OK it turns out Authentication does it that way. OK, maybe Logging actually the outlier here :).

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,121 @@
...
+                    duplicates.Add(check.Name);
+                }
+            }
+
+            if (duplicates.Count > 0)
+            {
+                throw new ArgumentException($"Duplicate health checks were registered with the name(s): {string.Join(", ", duplicates)}", nameof(healthChecks));
+            }
+
+            Checks = healthChecks.ToDictionary(c => c.Name);
```
Tratcher: Also needs to define the comparer for the dictionary

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthChecksBuilderAddCheckExtensions.cs
Tratcher: This comment seems to have been hidden without action or response.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        /// <summary>
+        /// Process an individual request.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public async Task InvokeAsync(HttpContext context)
+        {
+            if (context.Request.Path == _healthCheckOptions.Path)
```
rynowak: Should this be startswith? 

What about cases like `/health` vs `/health/` - should this be case-sensitive?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
rynowak: We might want to go further here and just provide a service that gets to generate the response. Since we plan on delivering integration packages for things like consul, we could ship it there and it could follow consul's conventions.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+                // TODO: Caching
+                // Get results
+                var result = await _healthCheckService.CheckHealthAsync(context.RequestAborted);
+
+                // Map status to response code
+                switch (result.Status)
+                {
+                    case HealthCheckStatus.Failed:
+                        // REVIEW: Maybe we shouldn't distinguish between failed health checks and unhealthy?
+                        context.Response.StatusCode = StatusCodes.Status500InternalServerError;
```
rynowak: I think we should. This will make it easier for us to support.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
rynowak: We should make 'responding' be a service and call it good.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                // Render results to JSON
+                // TODO: Authentication policy to control this? Sensitive data hiding? etc.
+                var json = new JObject(
+                    new JProperty("status", result.Status.ToString()),
+                    new JProperty("results", new JObject(result.Results.Select(pair =>
+                        new JProperty(pair.Key, new JObject(
+                            new JProperty("status", pair.Value.Status.ToString()),
+                            new JProperty("description", pair.Value.Description),
+                            new JProperty("data", new JObject(pair.Value.Data.Select(p => new JProperty(p.Key, p.Value))))))))));
```
rynowak: Should this be a `JsonSerializer` instead?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
Drawaes: Not only that it makes it easier to troubleshoot. Are my services going offline cause the coder sucks and wrote a bad healthcheck or is the service actually failing

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckResult.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        /// <summary>
+        /// Gets a human-readable description of the status of the component that was checked.
+        /// </summary>
+        public string Description => _description ?? string.Empty;
+
+        /// <summary>
+        /// Gets additional key-value pairs describing the health of the component.
+        /// </summary>
+        public IReadOnlyDictionary<string, object> Data => _data ?? _emptyReadOnlyDictionary;
```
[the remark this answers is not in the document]
anurse: The big reason to avoid string,string is that you may frequently want to emit numeric data (amount of memory consumed, etc.), and you want that preserved as a number in the JSON payload

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public static class HealthCheckAppBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a middleware that provides a REST API for requesting health check status.
+        /// </summary>
+        /// <param name="applicationBuilder">The <see cref="IApplicationBuilder"/></param>
+        /// <param name="path">The path on which to provide the API</param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseHealthChecks(this IApplicationBuilder applicationBuilder, PathString path)
```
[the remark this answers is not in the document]
anurse: I don't think the path belongs in ConfigureServices. The middleware and the service are not as tightly coupled as some of our other middleware-service pairs. The service is not tied to HTTP at all, it could be used to drive background monitoring tasks instead of pull-based HTTP endpoints.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: Would we need a service or would it just be configurations/modes for the middleware? Maybe `app.UseConsulHealthChecks()`, `app.UseKubernetesHealthChecks()`, etc. Or `app.UseHealthChecks(options => options.ConfigureForConsul())`, etc.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: I could be won over.

src/Microsoft.Extensions.Diagnostics.HealthChecks/CompositeHealthCheckResult.cs
Drawaes: Sometimes you don't have anything to check other than "can you ping the service" its often enough that you get a response.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: I'm legit unsure on this. I definitely thing we want `/health` and `/health/` to work, but I don't think we'd want `/health/foo` to work, since the user may want to map something else to that URL. For now, I'm going to change this to support `/health` and `/health/` and be case-insensitive. We can continue to discuss this though.

anurse: Hmm, looking at our other middleware, we generally do exact `==` checks for this. For example Authentication handlers for sign-in links: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs#L40

Of course, that's not exactly the same scenario here, since the Auth handlers also generate the urls and can make sure it's correct. Similarly, the Welcome page (not exactly an example of a "useful" middleware, I'll grant) also uses an exact path match:

https://github.com/aspnet/Diagnostics/blob/d390c369ffa7c0d6f83e61a6e01c3f42d8e6c5a0/src/Microsoft.AspNetCore.Diagnostics/WelcomePage/WelcomePageMiddleware.cs#L50

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckResult.cs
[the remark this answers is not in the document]
Drawaes: Yes! When a problem does occur. A structured error like a structured log is far more useful in automated systems to diagnose the issue

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: Yep, that's why I put it this way initially. I'll remove the comment, sounds like there's consensus :)

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -0,0 +1,121 @@
...
+                    try
+                    {
+                        _logger.LogTrace("Running health check: {healthCheckName}", pair.Key);
+                        result = await pair.Value.CheckHealthAsync(cancellationToken);
+                        _logger.LogTrace("Health check '{healthCheckName}' completed with status '{healthCheckStatus}'", pair.Key, result.Status);
+                    }
+                    catch (Exception ex)
+                    {
+                        // We don't log this as an error because a health check failing shouldn't bring down the active task.
+                        _logger.LogDebug(ex, "Health check '{healthCheckName}' threw an unexpected exception", pair.Key);
```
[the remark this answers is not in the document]
anurse: Fair. In that case would you expect it to throw?

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Runs all the health checks in the application and returns the aggregated status.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> which can be used to cancel the health checks.</param>
+        /// <returns>
+        /// A <see cref="Task{T}"/> which will complete when all the health checks have been run,
+        /// yielding a <see cref="CompositeHealthCheckResult"/> containing the results.
+        /// </returns>
+        Task<CompositeHealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
anurse: That sounds reasonable. Also makes it a little easier to look at adding things like groups/tags in a later version (though we should consider those before RTM as they may require breaking interface changes)

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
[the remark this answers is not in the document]
anurse: For now I'm making it log an error.

test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
```diff
@@ -0,0 +1,238 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests
+{
+    public class HealthCheckMiddlewareTests
```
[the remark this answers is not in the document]
anurse: Ok?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
Tratcher: == is overriden. https://github.com/aspnet/HttpAbstractions/blob/75a7552360cc1bfed1a56aad2180f730e5ba24cc/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs#L379

anurse:
Lots of :up: :date:s

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheck.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks
     /// A simple implementation of <see cref="IHealthCheck"/> which uses a provided delegate to
     /// implement the check.
     /// </summary>
-    public class HealthCheck : IHealthCheck
+    public sealed class HealthCheck : IHealthCheck
```
Tratcher: Since when have we used 'sealed' anywhere?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
Eilon: I agree that this should be "path equality", so not supporting `/health/foo` is good. Beyond that I don't super care how nitpicky we are about exact matches (casing, trailing slashes, etc.).

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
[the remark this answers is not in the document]
Tratcher: Why pass in the checks themselves rather than an `IEnumerable<string>` of check names? The service already has the check instances. Would you ever provide checks not currently registered with the service?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
Drawaes: Agreed... 99% of the time these are called by other services/ scripts anyway so exact matching is neither here nor there. Wild carding the whole path after Healy is a bad idea

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheck.cs
Drawaes: Only good reason is perf and devirtualisation. In that case if needed (probably not here) make a private sealed version internally of the public unsealed. Keeps it friendly and you get virtualisation. Public sealed types are nasty.

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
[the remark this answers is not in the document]
anurse: Allowing the caller to provide a list of instances saves dereferencing the dictionary, especially if the caller is managing a grouping of checks. But yeah, it could go either way.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Diagnostics.HealthChecks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class HealthCheckAppBuilderExtensions
```
Eilon: Add doc comments to all public stuff. For this one do something like: https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageExtensions.cs#L11

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Adds a middleware that provides a REST API for requesting health check status.
+        /// </summary>
+        /// <param name="applicationBuilder">The <see cref="IApplicationBuilder"/></param>
+        /// <param name="path">The path on which to provide the API</param>
+        /// <returns></returns>
```
Eilon: Missing doc comment here. Example: https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageExtensions.cs#L19

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class HealthCheckAppBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a middleware that provides a REST API for requesting health check status.
+        /// </summary>
+        /// <param name="applicationBuilder">The <see cref="IApplicationBuilder"/></param>
+        /// <param name="path">The path on which to provide the API</param>
```
Eilon: And end these sentences in periods 😄

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
anurse: As far as I can tell, we don't do whatever @Eilon is calling "path equality" (I assume you mean treating `/health` and `/health/` the same) anywhere else in our stack, and the equality operator for `PathString` doesn't do it. If we want to add logic to do that, I can look in to that, but I don't think it's necessary, since as @Drawaes said these are usually just punched into an automated monitoring tool.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheck.cs
anurse: There's no place for virtualization here, since this is basically just a holder of a name-func pair. If people want to avoid `sealed` here though I'm not opposed.

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Abstractions for defining health checks in .NET applications</Description>
```
Eilon: And in a period. And/or add some "commonly used types," a la https://github.com/aspnet/EntityFrameworkCore/blob/e3e76e10ceed8a0610e05fa3f787a2470421abe5/src/EFCore/EFCore.csproj#L6-L8

Eilon (approved):
Looks great. Just some comments about doc comments and other metadata thingies.

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckResult.cs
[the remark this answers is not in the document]
anurse: > Or we can keep it <string, object> and just be ready to say no to those requests when they come.

Yeah, that's my approach. You want to serialize it, you can build your own middleware ;)

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
anurse:
```csharp
/// <summary>
/// I love doc comments, they're the best
/// </summary>
```

anurse:
I'd like to close this out soon. We can continue iterating with additional issues if any still exist. I'll merge tonight if there are no other requests for changes.

Drawaes:
LGTM

Drawaes (approved):

Eilon:
I personally think this is good to go. Once we and the community get into writing a whole heckuva lot of health checks, we'll find what the *real* issues are (now that the doc comments are fixed! 😄 ).

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
rynowak: Ideally `UseHealthChecks()` does everything for you, and you if you want to control it you can use the options.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
rynowak: Cool beans, I just wanted to make sure it was thought through.

rynowak (approved):

Issue #400, opened 2017-07-29
ExceptionHandlerMiddleware uses _next as default ExceptionHandler?!


NinoFloris:
EDIT: The story is even weirder...

ExceptionHandlerMiddleware uses _next as default ExceptionHandler.
So let's say my aspnet controller throws an exception. this is bubbled and caught in the first try catch of the ExceptionHandlerMiddleware, then because most people will just run with the default `app.UseExceptionHandler()` we haven't explicitly set an ExceptionHandler and so it got set to `_next` by the constructor. This means we'll pointlessly run the same pipeline again in the second inner try catch: https://github.com/aspnet/Diagnostics/blob/4e044a1e30454b87edbc316f40ba608d1160cb28/src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L74
this will of course result in the same exception 99% of the time.

And it also means most people will see two error log messages as for most of the cases (when `ExceptionHandler = _next`) will also log:
```cs
_logger.LogError(0, ex2, "An exception was thrown attempting to execute the error handler.");
```

Workaround:
```cs
var opts = new ExceptionHandlerOptions{ ExceptionHandler = ctx => Task.CompletedTask };
app.UseExceptionHandler(opts) 
```

This default is crazy, a waste of resources, and how this has never been reported is a mystery to me.

Tratcher:
ExceptionHandler defaults to _next for the ExceptionHandlingPath scenario (though we should add a check in the constructor that ExceptionHandlingPath is actually set). This scenario is the one used in templates (`UseExceptionHandler("/error")`). I haven't seen the other scenarios used in the wild.

davidfowl:
Why do we have that overload at all?

Tratcher:
Which overload, the no params one? For DI.

Eilon:
I believe this is what @Tratcher is talking about: There's a UseExceptionHandler overload that takes no params (aside from `this`): https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerExtensions.cs#L19

[...]

PR #417, opened 2017-10-27
UseExceptionHandler throws if ExceptionHandlingPath not set

jbagga:
Addresses #400 

cc @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics/Resources.resx
+++ b/src/Microsoft.AspNetCore.Diagnostics/Resources.resx
@@ -247,4 +247,7 @@
   <data name="RuntimeInfoPage_Environment" xml:space="preserve">
     <value>Environment:</value>
   </data>
+  <data name="ExceptionHandlerOptions_NotConfiguredCorrectly" xml:space="preserve">
+    <value>The exception handler middleware options 'ExceptionHandlingPath' and 'ExceptionHandler' are both null. Please set either both or at least the 'ExceptionHandlingPath' when using 'app.UseExceptionHandler()'.</value>
+  </data>
 </root>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Diagnostics/Resources.resx
```diff
@@ -247,4 +247,7 @@
   <data name="RuntimeInfoPage_Environment" xml:space="preserve">
     <value>Environment:</value>
   </data>
+  <data name="ExceptionHandlerOptions_NotConfiguredCorrectly" xml:space="preserve">
+    <value>The exception handler middleware options 'ExceptionHandlingPath' and 'ExceptionHandler' are both null. Please set either both or at least the 'ExceptionHandlingPath' when using 'app.UseExceptionHandler()'.</value>
+  </data>
```
jbagga: Perhaps too verbose?

Tratcher: Agreed, you can combine those to say you must set either the path or the handler.

test/Microsoft.AspNetCore.Diagnostics.Tests/ExceptionHandlerTest.cs
```diff
@@ -494,5 +494,26 @@ public class ExceptionHandlerTest
...
+            // Arrange
+            DiagnosticListener diagnosticListener = null;
+
+            var builder = new WebHostBuilder()
+                .Configure(app =>
+                {
+                    diagnosticListener = app.ApplicationServices.GetRequiredService<DiagnosticListener>();
+
+                    // Act
+                    var exception = Assert.Throws<InvalidOperationException>(() => app.UseExceptionHandler());
```
Tratcher: The exception won't be thrown until you call builder.Build().

jbagga: No, the exception is thrown from the ctor. Unless that isn't what should happen

PR #431, opened 2018-01-09
Show correct missing key name in NoContextType error message

andrewjsaid:
The `MigrationsEndPointMiddleware` class (/dev/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/MigrationsEndPointMiddleware.cs) requires a "context" form value in the form-data, but if missing complains with the error message:

> No context type was specified. Ensure the form data from the request includes a contextTypeName value, specifying the context to apply migrations for.

This error message is misleading as one would expect that the request key must be "contextTypeName". I only figured out what it must be by exploring the source code.

This commit changes the message to display a clearer error message; specifically mentioning the context key as follows:
> No context type was specified. Ensure the form data from the request includes a context value, specifying the context type name to apply migrations for.

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/Internal/DiagnosticsEntityFrameworkCoreLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/Internal/DiagnosticsEntityFrameworkCoreLoggerExtensions.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.Internal
         private static readonly Action<ILogger, Exception> _noContextType = LoggerMessage.Define(
             LogLevel.Error,
             new EventId(1, "NoContextType"),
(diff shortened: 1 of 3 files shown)


Eilon:
And cc @ajcvickers

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/Internal/DiagnosticsEntityFrameworkCoreLoggerExtensions.cs
```diff
@@ -12,7 +12,7 @@ internal static class DiagnosticsEntityFrameworkCoreLoggerExtensions
         private static readonly Action<ILogger, Exception> _noContextType = LoggerMessage.Define(
             LogLevel.Error,
             new EventId(1, "NoContextType"),
-            "No context type was specified. Ensure the form data from the request includes a contextTypeName value, specifying the context to apply migrations for.");
+            "No context type was specified. Ensure the form data from the request includes a context value, specifying the context type name to apply migrations for.");
```
anpete: 'context' for clarity

pranavkm:
cc @ajcvickers

PR #433, opened 2018-01-26
Fixes null ref in DatabaseErrorPageMiddleware.

anpete:
Fix aspnet/Home#2820 - UseDatabaseErrorPage throws NullReferenceException when ef exception occur in background task
Fix aspnet/EntityFrameworkCore#9599 - Insert entity in non existing table throws NullException

Added null checks to async local access in event callback.

Related to aspnet/Home#2825

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
@@ -78,6 +78,8 @@ namespace Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore
             _options = options.Value;
             _logger = loggerFactory.CreateLogger<DatabaseErrorPageMiddleware>();
 
+            // Note: this currently leaks if the server hosting this middleware is disposed.
+            // See aspnet/Home #2825
             DiagnosticListener.AllListeners.Subscribe(this);
         }
 
@@ -121,7 +123,6 @@ namespace Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore
                         else
                         {
                             var relationalDatabaseCreator = context.GetService<IDatabaseCreator>() as IRelationalDatabaseCreator;
-
                             if (relationalDatabaseCreator == null)
                             {
                                 _logger.NotRelationalDatabase();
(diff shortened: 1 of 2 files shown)


ajcvickers (approved):

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
```diff
@@ -216,15 +217,17 @@ void IObserver<DiagnosticListener>.OnNext(DiagnosticListener diagnosticListener)
         {
             switch (keyValuePair.Value)
             {
+                // NB: _localDiagnostic.Value can be null when this middleware has been leaked.
```
Tratcher: can be null when DB operations take place outside of a request thread.

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
```diff
@@ -78,6 +78,8 @@ public void Hold(Exception exception, Type contextType)
             _options = options.Value;
             _logger = loggerFactory.CreateLogger<DatabaseErrorPageMiddleware>();
 
+            // Note: this currently leaks if the server hosting this middleware is disposed.
```
Tratcher: Want to try the IMiddleware fix here?

Tratcher (commented):
You opted not to try simplifying the design instead?

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
anpete: Depends what it is :trollface:

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
anpete: Want me to add that?

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
Tratcher: Change the class to implement IMiddleware and IDisposable. The catch is that the class has to be registered as a service. Diagnostics does not currently register any services.

anpete:
Re the design simplification, it turns out that any random database exception can end up here, and we need a way to deduce that it came from EF. Would love to improve this if you have any ideas.

Tratcher:
Don't they all get wrapped?

anpete:
No.

src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/DatabaseErrorPageMiddleware.cs
anpete: What does registration of the middleware look like from UseDatabaseErrorPage?

Tratcher: It's still `app.UseMiddleware<DatabaseErrorPageMiddleware>`, but passing in options directly might not work.

PR #438, opened 2018-02-15
Use ExperimentalPackageVersion

ryanbrandenburg:


version.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>2.1.0</VersionPrefix>
-    <ExperimentalProjectVersionPrefix>0.4.0</ExperimentalProjectVersionPrefix>
+    <ExperimentalVersionPrefix>0.4.0</ExperimentalVersionPrefix><ExperimentalPackageVersion Condition="'$(IsFinalBuild)' == 'true'">$(ExperimentalVersionPrefix)-$(ExperimentalVersionSuffix)-final</ExperimentalPackageVersion>
```
natemcmaster: nit: new line.

Also add ExperimentalVersionSuffix

version.props
```diff
@@ -1,7 +1,9 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>2.1.0</VersionPrefix>
-    <ExperimentalVersionPrefix>0.4.0</ExperimentalVersionPrefix><ExperimentalPackageVersion Condition="'$(IsFinalBuild)' == 'true'">$(ExperimentalVersionPrefix)-$(ExperimentalVersionSuffix)-final</ExperimentalPackageVersion>
+    <ExperimentalVersionPrefix>0.4.0</ExperimentalVersionPrefix>
+    <ExperimentalVersionSuffix>alpha1</ExperimentalVersionSuffix>
+    <ExperimentalPackageVersion Condition="'$(IsFinalBuild)' == 'true'">$(ExperimentalVersionPrefix)-$(ExperimentalVersionSuffix)-final</ExperimentalPackageVersion>
```
natemcmaster: We don't need this now, but for RTM we'll need to check if `'$(ExperimentalVersionSuffix)' == 'rtm' if we want to produce just `0.4.0` as the package version.

src/Microsoft.AspNetCore.Diagnostics.Elm/Microsoft.AspNetCore.Diagnostics.Elm.csproj
```diff
@@ -2,8 +2,10 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core Error Logging Middleware (ELM) to capture and display request logs.</Description>
-    <VersionPrefix Condition="'$(ExperimentalProjectVersionPrefix)' != ''">$(ExperimentalProjectVersionPrefix)</VersionPrefix>
-    <VerifyVersion Condition="'$(ExperimentalProjectVersionPrefix)' != ''">false</VerifyVersion>
+    <VersionPrefix Condition="'$(ExperimentalVersionPrefix)' != ''">$(ExperimentalProjectVersionPrefix)</VersionPrefix>
```
pranavkm: Can we remove the conditional check?

natemcmaster: I used the pattern intentionally. If someone were to unset ExperimentalVersionPrefix, NuGet will default to producing a package version "1.0.0". I figured it was better to have it fallback to the VersionPrefix in version.props than to produce "1.0.0"

PR #441, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


src/Microsoft.AspNetCore.Diagnostics.Elm/baseline.netcore.json
```diff
@@ -1,417 +1,510 @@
 {
-  "AssemblyIdentity": "Microsoft.AspNetCore.Diagnostics.Elm, Version=0.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
+  "AssemblyIdentity": "Microsoft.AspNetCore.Diagnostics.Elm, Version=0.3.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
```
Tratcher: @javiercn we should exclude prototypes from having / requiring baselines.

javiercn: Filed https://github.com/aspnet/BuildTools/issues/614 for it. /cc: @Eilon

PR #445, opened 2018-06-07
Adding VSTS file

ryanbrandenburg:


ryanbrandenburg:
It appears that tests fail on VSTS due to timeouts. I need a SME to look into it, @ajaybhargavb is that you? If not I have no clue who as most of the commits to this repo are automated.

ajaybhargavb:
I looked at the failure and unfortunately I have no clue what it's about. Maybe @kichalla could help.

ryanbrandenburg:
@ajcvickers can we get someone from the EF team to look at this? It seems to be issues related to database logins and whatnot. Is there a prerequisite we're maybe missing (if so this scenario should make better error messages)?

ajcvickers:
@smitpatel Can you take a look at this?

smitpatel:
The possible root cause of failures here, localdb instance is not running when testing starts so initial tests fail. But once the instance is running, all tests would start passing.

ajcvickers:
We have hit issues similar to this in the past when the machine/VM that is running the tests is too slow for SQL Server to startup in a reasonable time. Do we know what kind of machine is being used here?

ryanbrandenburg:
Judging by other issues we've been running into I'm assuming the machines are pretty slow and possibly single-core. This is to be expected since we're using the lowest tier of hosted machines for now, but I also see it as an opportunity to ensure our product/test/build works on more modest setups unlike the beefy machines we usually use to test.

When this has happened in the past what was the resolution? Should we poll for SqlServer being in a ready state, or is there a "launch SqlServer and only return once it's in a ready state" command that we can run before build.cmd|sh?

smitpatel:
`sqllocaldb start` command should start the default instance (which we use in testing anyway). Unsure if the command is in the path. Can you try it out if its easy to test?

ajcvickers:
@ryanbrandenburg In the past the resolution has been to use faster machines. It's a laudable goal to say we should be able to run our tests on modest setups,, but if one of the main things we depend on--that is, SQL Server Localdb--cannot run reliably on modest setups then to achieve this goal we would need to move away from using that dependency. That is something we have considered--for example, using SQLite for most testing--but that obviously impacts many things, so the motivation for doing so would have to be very strong.

Alternately, if the @smitpatel's idea pans out, then that might help.

.vsts-pipelines/builds/ci-public.yml
```diff
@@ -12,4 +12,23 @@ resources:
...
+    agentOs: Windows
+    beforeBuild:
+    - script: sqllocaldb start
+      displayName: Start LocalDB
+
+- template: .vsts-pipelines/templates/phases/default-build.yml@buildtools
+  parameters:
+    agentOs: macOS
+    beforeBuild:
+    - script: sqllocaldb start
```
smitpatel: macOS/Linux don't run SqlServer tests.

ryanbrandenburg:
Looks like @smitpatel's suggestion took care of it. The remaining failures are problems with dependencies.

PR #454, opened 2018-07-31
Updating Health Checks for 2.2

rynowak:
A bunch of small changes and updates for 2.2 focused at making our main
scenarios more streamlined and focused. Also adds samples for
extensibility we support so far.

A list of changes:

Clearing baselines for these projects. We didn't ship anything in 2.1 so
there should be nothing in the baselines.

--

The middleware now uses Map for path matching. This makes the actual
`HealthCheckMiddleware` more standalone. This will make it easy to use
with Dispatcher/Endpoint Routing in the future.

This also manifests by removing Path from HealthCheckOptions - the path
is an explicit argument to the UseHealthChecks middelware - this
streamlines the design for 3.0.

--

Added extensibility for customizing the status codes (aspnet/Home#2584)

--

Added extensibility for writing the textual output (aspnet/Home#2583)

--

Changed the default output to be `text/plain`. The most common use cases
for health checks don't include a detailed status.

The existing output format is still available as an option.


src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckResponseWriters.cs
```diff
@@ -0,0 +1,56 @@
...
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            if (result == null)
+            {
+                throw new ArgumentNullException(nameof(result));
+            }
+
+            httpContext.Response.ContentType = "text/plain";
+            return httpContext.Response.WriteAsync(result.Status.ToString());
```
pranavkm: `InvariantCulture`

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -19,60 +15,62 @@ public class HealthCheckMiddleware
...
-                            new JProperty("description", pair.Value.Description),
-                            new JProperty("data", new JObject(pair.Value.Data.Select(p => new JProperty(p.Key, p.Value))))))))));
-                await context.Response.WriteAsync(json.ToString(Formatting.None));
+            // Map status to response code - this is customizable via options. 
+            if (!_healthCheckOptions.ResultStatusCodes.TryGetValue(result.Status, out int statusCode))
+            {
+                var message =
+                    $"No status code mapping found for {nameof(HealthCheckStatus)} value: {result.Status}." +
+                    $"{nameof(HealthCheckOptions)}.{nameof(HealthCheckOptions.ResultStatusCodes)} must contain" +
+                    $"and entry for {result.Status}.";
```
pranavkm: > an~~d~~ entry

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -19,60 +15,62 @@ public class HealthCheckMiddleware
...
-                var json = new JObject(
-                    new JProperty("status", result.Status.ToString()),
-                    new JProperty("results", new JObject(result.Results.Select(pair =>
-                        new JProperty(pair.Key, new JObject(
-                            new JProperty("status", pair.Value.Status.ToString()),
-                            new JProperty("description", pair.Value.Description),
-                            new JProperty("data", new JObject(pair.Value.Data.Select(p => new JProperty(p.Key, p.Value))))))))));
-                await context.Response.WriteAsync(json.ToString(Formatting.None));
+            // Map status to response code - this is customizable via options. 
+            if (!_healthCheckOptions.ResultStatusCodes.TryGetValue(result.Status, out int statusCode))
```
pranavkm: `out var` (all day everyday)

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -14,19 +14,54 @@ namespace Microsoft.AspNetCore.Builder
...
         public static IApplicationBuilder UseHealthChecks(this IApplicationBuilder app, PathString path)
         {
-            app = app ?? throw new ArgumentNullException(nameof(app));
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
 
-            return app.UseMiddleware<HealthCheckMiddleware>(Options.Create(new HealthCheckOptions()
+            if (string.IsNullOrEmpty(path.Value))
```
pranavkm: `if (!path.HasValue)`

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
```diff
@@ -14,19 +14,54 @@ namespace Microsoft.AspNetCore.Builder
...
             {
-                Path = path
-            }));
+                throw new ArgumentException("A URL path must be provided", nameof(path));
```
pranavkm: > Request path must be provided

"URL path" might be confusing.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckOptions.cs
```diff
@@ -10,9 +14,23 @@ namespace Microsoft.AspNetCore.Diagnostics.HealthChecks
...
         /// <summary>
-        /// Gets or sets the path at which the Health Check results will be available.
+        /// Gets or sets a delegate used to write the response.
         /// </summary>
-        public PathString Path { get; set; }
+        /// <remarks>
+        /// The default value is a delegate that will write a minimal <c>text/plain</c> response with the value
+        /// of <see cref="CompositeHealthCheckResult.Status"/> as a string.
+        /// </remarks>
+        public Func<HttpContext, CompositeHealthCheckResult, Task> ResponseWriter { get; set; } = HealthCheckResponseWriters.WriteMinimalPlaintext;
```
pranavkm: If we expect users to author these, should this be an interface?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckResponseWriters.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            if (result == null)
+            {
+                throw new ArgumentNullException(nameof(result));
+            }
+
+            httpContext.Response.ContentType = "application/json";
+
+            var json = new JObject(
```
pranavkm: Not for this PR, but maybe something we should consider as a format  - https://inadarei.github.io/rfc-healthcheck/#rfc.section.5

pranavkm (approved):

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckResponseWriters.cs
rynowak: https://msdn.microsoft.com/en-us/library/3bkcbx85(v=vs.110).aspx - they actually want you to use the no-arg version.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
rynowak: In what way? I don't get how URL path is confusing in the context of URLS

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckOptions.cs
rynowak: We're specifically not introducing an abstraction here. When/if someday we have formatters at this level of the stack, this delegate will be used to plug that in.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckResponseWriters.cs
rynowak: It's been discussed, in general we're not that interested in defining a format for folks to use for health checks. We might even remove this code because I'm not sure we'd tell users to use it.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckAppBuilderExtensions.cs
pranavkm: It sounds like a regular url (with protocol and port). But I guess creating a `PathString` would complain about that

rynowak: it says url path.

PR #455, opened 2018-08-01
Add filtering to Health Checks middleware

rynowak:
This allows each middleware to be configured with a specific set of
checks (by name). See the comments in the sample for how this is
frequently used.

This is also addresses aspnet/Home#2575 - or at least the part that we
plan to do. We think that any sort of built-in system of metadata or
tags is vast overkill, and doesn't really align with the primary usage
of health checks.

We're providing building blocks, and these can be put together to
build more complicated things like what's described in aspnet/Home#2575.

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -73,5 +78,41 @@ public class HealthCheckMiddleware
...
+                if (!notFound.Remove(kvp.Key))
+                {
+                    // This check was excluded
+                    continue;
+                }
+
+                matches.Add(kvp.Value);
+            }
+
+            if (notFound.Count >0)
```
pranavkm: formatting

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -73,5 +78,41 @@ public class HealthCheckMiddleware
...
+            {
+                var message = 
+                    $"The following health checks were not found: '{string.Join(", ", notFound)}'." +
```
pranavkm: Needs a space after the dot.

test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
```diff
@@ -346,5 +346,68 @@ public class HealthCheckMiddlewareTests
...
+                            "Bazzzzzz",
+                            "FOO",
+                        },
+                    });
+                })
+                .ConfigureServices(services =>
+                {
+                    services.AddHealthChecks()
+                        .AddCheck("Foo", () => Task.FromResult(HealthCheckResult.Healthy("A-ok!")))
+                        // Will get filtered out
```
pranavkm: Remove

PR #466, opened 2018-08-23
Allow health checks to use any DI lifetime

rynowak:
This change allows registered IHealthCheck implementations to use any DI
lifetime. This is necessary for scenarios like using EF which requires a
scope.

The works by having the health check service create a scope for each
time it queries health checks. This scope does not overlap or share
state with other scopes (the request scope) so there is no crosstalk
between processing going on per-request in ASP.NET Core and the health
check operation.


samples/HealthChecksSample/LivenessProbeStartup.cs
```diff
@@ -53,11 +52,8 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             // The liveness check uses an 'identity' health check that always returns healty
             app.UseHealthChecks("/health/live", new HealthCheckOptions()
             {
-                // Filters the set of health checks run by this middleware
-                HealthCheckNames =
-                {
-                    "identity",
-                },
+                // Exclude all checks, just return a 200.
+                Predicate = (check) => false,
```
rynowak: Changing this to an arbitrary predicate rather than an explicit 'filter by name' seems like a good usability improvement. Since the set of checks has to be resolved per-request a lot of motivations for providing 'filter by name' seemed to melt away.

src/Microsoft.Extensions.Diagnostics.HealthChecks/IHealthCheckService.cs
```diff
@@ -1,26 +1,38 @@
...
     public interface IHealthCheckService
     {
-        /// <summary>
-        /// A <see cref="IReadOnlyDictionary{TKey, T}"/> containing all the health checks registered in the application.
-        /// </summary>
-        /// <remarks>
-        /// The key maps to the <see cref="IHealthCheck.Name"/> property of the health check, and the value is the <see cref="IHealthCheck"/>
-        /// instance itself.
-        /// </remarks>
-        IReadOnlyDictionary<string, IHealthCheck> Checks { get; }
```
rynowak: This has to go away now because they aren't singletons. The main use case we had for providing this was to support filtering. If someone wants to get the set of checks for an arbitrary reason they can create a scope and resolve them 😆

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthChecksBuilderAddCheckExtensions.cs
```diff
@@ -88,5 +88,32 @@ public static IHealthChecksBuilder AddCheck(this IHealthChecksBuilder builder, I
...
+        /// check allows you to register a health check that depends on transient and scoped services.
+        /// </remarks>
+        public static IHealthChecksBuilder AddCheck<T>(this IHealthChecksBuilder builder) where T : class
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (typeof(T).IsAssignableFrom(typeof(IHealthCheck)))
```
pranavkm: `typeof(IHealthCheck).IsAssignableFrom(typeof(T))`

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthChecksBuilderAddCheckExtensions.cs
```diff
@@ -88,5 +88,32 @@ public static IHealthChecksBuilder AddCheck(this IHealthChecksBuilder builder, I
...
+        /// </summary>
+        /// <typeparam name="T">The health check implementation type.</typeparam>
+        /// <param name="builder">The <see cref="IHealthChecksBuilder"/>.</param>
+        /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
+        /// <remarks>
+        /// This method will register a transient service of type <see cref="IHealthCheck"/> with the 
+        /// provided implementation type <typeparamref name="T"/>. Using this method to register a health
+        /// check allows you to register a health check that depends on transient and scoped services.
+        /// </remarks>
+        public static IHealthChecksBuilder AddCheck<T>(this IHealthChecksBuilder builder) where T : class
```
pranavkm: Couldn't you do `where T : class, IHealthCheck`? Better than doing a runtime check

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -6,127 +6,103 @@
...
+            using (var scope = _scopeFactory.CreateScope())
             {
-                throw new ArgumentException($"Duplicate health checks were registered with the name(s): {string.Join(", ", duplicates)}", nameof(healthChecks));
-            }
+                var healthChecks = scope.ServiceProvider.GetRequiredService<IEnumerable<IHealthCheck>>();
 
-            Checks = healthChecks.ToDictionary(c => c.Name, StringComparer.OrdinalIgnoreCase);
+                if (predicate != null)
+                {
+                    healthChecks = healthChecks.Where(predicate);
```
pranavkm: You could avoid the Linq call by doing this inside the `foreach`

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -6,127 +6,103 @@
...
+            // Scan the list for duplicate names to provide a better error if there are duplicates.
+            var names = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var duplicates = new List<string>();
+            foreach (var check in healthChecks)
             {
-                cancellationToken.ThrowIfCancellationRequested();
-                // If the health check does things like make Database queries using EF or backend HTTP calls,
-                // it may be valuable to know that logs it generates are part of a health check. So we start a scope.
-                using (_logger.BeginScope(new HealthCheckLogScope(check.Name)))
+                if (!names.Add(check.Name))
```
pranavkm: This would print the same name multiple times if the check is registered more than twice. Think a `GroupBy` would be terse and more correct

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckOptions.cs
```diff
@@ -15,15 +15,19 @@ namespace Microsoft.AspNetCore.Diagnostics.HealthChecks
...
         /// </summary>
         /// <remarks>
-        /// If <see cref="HealthCheckNames"/> is empty, the <see cref="HealthCheckMiddleware"/> will run all
+        /// If <see cref="Predicate"/> is <c>null</c>, the <see cref="HealthCheckMiddleware"/> will run all
         /// registered health checks - this is the default behavior. To run a subset of health checks,
-        /// add the names of the desired health checks.
+        /// provide a function that filters the set of checks.
         /// </remarks>
-        public ISet<string> HealthCheckNames { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        public Func<IHealthCheck, bool> Predicate { get; set; }
```
davidfowl: Filter?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -6,127 +6,103 @@
...
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    // If the health check does things like make Database queries using EF or backend HTTP calls,
+                    // it may be valuable to know that logs it generates are part of a health check. So we start a scope.
+                    using (_logger.BeginScope(new HealthCheckLogScope(healthCheck.Name)))
+                    {
+                        HealthCheckResult result;
+                        try
+                        {
+                            _logger.LogTrace("Running health check: {healthCheckName}", healthCheck.Name);
```
davidfowl: This isn't the logging pattern we use in components.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthChecksBuilderAddCheckExtensions.cs
rynowak: yeah going point. I think this was a bad copy-paste.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
rynowak: sure

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
rynowak: sure good point.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckOptions.cs
rynowak: Eh, we considered it and thought this was less overloaded. If we use filter (or another term) other places in the framework then I want to change this.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
rynowak: YEAH BRO

rynowak:
Updated this and cleaned up the logging. I logged an issue to consider the logging further, I don't think what's there is good enough but I don't want to drill into as part of fixing DI.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckService.cs
```diff
@@ -6,127 +6,139 @@
...
-                        result = await check.CheckHealthAsync(cancellationToken);
-                        _logger.LogTrace("Health check '{healthCheckName}' completed with status '{healthCheckStatus}'", check.Name, result.Status);
-                    }
-                    catch (Exception ex)
-                    {
-                        // We don't log this as an error because a health check failing shouldn't bring down the active task.
-                        _logger.LogError(ex, "Health check '{healthCheckName}' threw an unexpected exception", check.Name);
-                        result = new HealthCheckResult(HealthCheckStatus.Failed, ex, ex.Message, data: null);
-                    }
+                public static readonly EventId HealthCheckBegin = new EventId(100, "HealthCheckBegin");
```
pranavkm: Nice

PR #479, opened 2018-09-16
Use options for registering health checks

rynowak:
This change pivots to use options for registering health checks. We get
a few pretty nice things out of this, and it unblocks some of our
requirements.

Now all registration methods support the application developer
configuring the name, failure-status, and tags for each health check.
This is a requirment, that we weren't really satisfying - which is what
led to this redesign. In support of this health checks now return pass/fail,
and the service is responsible for assigning the status.

----

Health check authors that need configuration data (connection string as
an example) now have three ways to do this depending on their
requirements.
1. Create an instance and register that (easiest)
2. Use Type Activation and pass parameters (middle)
3. Use named options (richest)

We expect most health checks to need/want some kind of configuration -
which 1) works pretty well to solve. However many other health checks
will need DI + configuration. It was also a gap that we didn't have a
good way to use named options, when it's such a good fit for our
scenarios.

Added new registration methods for type activation that allow you to
pass parameters for 2).

Added a context type that allows the running health check access to its
registration for 3).

----

Redesigned and renamed how status gets reported. Health checks return
pass/fail result, but the overall HealthReport includes entries of a
different type. This seems fine because there isn't really a way to
consume a HealthCheckResult directly - the service is the only consumer.

----

Added support for tags. This was easy to add now that we have a separate
registration type, and it's quite handy for building filters (see
sample).


samples/HealthChecksSample/GCInfoHealthCheck.cs
```diff
@@ -2,21 +2,53 @@
...
 namespace HealthChecksSample
 {
     // This is an example of a custom health check that implements IHealthCheck.
     //
-    // See CustomWriterStartup to see how this is registered.
+    // This example also shows a technique for authoring a health check that needs to be registered
+    // with additional configuration data. This technique works via named options, and is useful
+    // for authoring health checks that can be disctributed as libraries.
+    
+    public static class GCInfoHealthCheckBuilderExtensions
```
rynowak: This is a sample that hews pretty close to what we'd tell a library author to do if they want to ship a health check as a package.

samples/HealthChecksSample/GCInfoHealthCheck.cs
```diff
@@ -2,21 +2,58 @@
...
+    {
+        public static IHealthChecksBuilder AddGCInfoCheck(
+            this IHealthChecksBuilder builder, 
+            string name, 
+            HealthStatus? failureStatus = null, 
+            IEnumerable<string> tags = null, 
+            long? thresholdInBytes = null)
+        {
```
rynowak: I'm tempting to put some kind of `Action<HealthCheckRegistration>` in here, but it doesn't play super nice with all of these signatures.

samples/HealthChecksSample/LivenessProbeStartup.cs
```diff
@@ -17,7 +15,7 @@ public void ConfigureServices(IServiceCollection services)
             // Registers required services for health checks
             services
                 .AddHealthChecks()
-                .AddCheck(new SlowDependencyHealthCheck());
+                .AddCheck<SlowDependencyHealthCheck>("Slow", failureStatus: null, tags: new[] { "ready", });
```
rynowak: We could consider reordering some parameters to make these signatures better for optional parameters. What isn't great is the ones that take a delegate, because its natural to put the delegate at the end. Thoughts?

src/Microsoft.Extensions.Diagnostics.HealthChecks/DependencyInjection/HealthChecksBuilderDelegateExtensions.cs
```diff
@@ -0,0 +1,417 @@
...
+        /// </summary>
+        /// <param name="builder">The <see cref="IHealthChecksBuilder"/>.</param>
+        /// <param name="name">The name of the health check.</param>
+        /// <param name="failureStatus">
+        /// The <see cref="HealthStatus"/> that should be reported when the health check reports a failure. If the provided value
+        /// is <c>null</c>, then <see cref="HealthStatus.Unhealthy"/> will be reported.
+        /// </param>
+        /// <param name="check">A delegate that provides the health check implementation.</param>
+        /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
+        public static IHealthChecksBuilder AddDelegateCheck(
```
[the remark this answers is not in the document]
rynowak: What would you prefer? I thought `AddCheck` was becoming way too overloaded. This is our builder type, so if we're not adding methods with different names then we're ignoring a key element of the design space.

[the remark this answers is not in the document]
davidfowl: AddCheck and AddAsyncCheck seem fine no? What does the intellisense look like?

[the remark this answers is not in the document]
rynowak: `AddCheck` is the same as most of these: https://github.com/aspnet/Diagnostics/blob/2cbdb2c428f7ce5eb3c7241e897233882352b8a5/src/Microsoft.Extensions.Diagnostics.HealthChecks/DependencyInjection/HealthChecksBuilderAddCheckExtensions.cs

That's what I'm trying to avoid in case we end up adding more stuff to those.

samples/HealthChecksSample/GCInfoHealthCheck.cs
```diff
@@ -2,21 +2,58 @@
...
+            // Register a check of type GCInfo
+            builder.AddCheck<GCInfoHealthCheck>(name, failureStatus ?? HealthStatus.Degraded, tags);
+
+            // Configure named options to pass the threshold into the check.
+            if (thresholdInBytes.HasValue)
+            {
+                builder.Services.Configure<GCInfoOptions>(options =>
```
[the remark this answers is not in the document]
rynowak: ya, you're right this is a bug.

rynowak:
@davidfowl updates

src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
```diff
@@ -0,0 +1,149 @@
...
+
+                            entry = new HealthReportEntry(
+                                result.Result ? HealthStatus.Healthy : registration.FailureStatus,
+                                result.Description,
+                                result.Exception,
+                                result.Data);
+
+                            Log.HealthCheckEnd(_logger, registration, entry, stopwatch.GetElapsedTime());
+                        }
+                        catch (Exception ex)
```
davidfowl: Do you want to handle OperationCancelledException differently?

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckResult.cs
```diff
@@ -13,176 +13,65 @@ public struct HealthCheckResult
...
         /// </summary>
-        /// <returns>A <see cref="HealthCheckResult"/> representing a component in a degraded state.</returns>
-        /// <param name="description">A human-readable description of the status of the component that was checked.</param>
-        /// <param name="exception">An <see cref="Exception"/> representing the exception that was thrown when checking for status (if any).</param>
-        public static HealthCheckResult Degraded(string description, Exception exception)
-            => new HealthCheckResult(HealthCheckStatus.Degraded, exception, description, data: null);
+        /// <returns>A <see cref="HealthCheckResult"/> representing a passing component.</returns>
+        /// <param name="description">A human-readable description of the status of the component that was checked. Optional.</param>
+        /// <param name="data">Additional key-value pairs describing the health of the component. Optional.</param>
+        public static HealthCheckResult Passed(string description = null, IReadOnlyDictionary<string, object> data = null)
```
davidfowl: Success instead of Passed

rynowak: Passed/Failed vs Successed/Failed?

src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
rynowak: Yeah, we have an issue for that https://github.com/aspnet/Diagnostics/issues/480

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/HealthCheckResult.cs
rynowak: you'll understand in 5 yrs

steveoh:
Are health checks all pass/fail in reality? 

I think there is an in-between place since we can use healthy, **degraded**, and unhealthy. If you limit the check to pass/fail and then use the `failStatus` to mark the check as degraded or unhealthy, it seems that you lose the ability to choose degraded _or_ unhealthy. Am I missing something?

rynowak:
Thanks for this feedback, I'm opening an issue to track this so we can take a look into changing it.

PR #482, opened 2018-09-20
Allow cancellation to propagate

rynowak:


src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
```diff
@@ -77,7 +77,9 @@ internal class DefaultHealthCheckService : HealthCheckService
 
                             Log.HealthCheckEnd(_logger, registration, entry, stopwatch.GetElapsedTime());
                         }
-                        catch (Exception ex)
+
+                        // Allow cancellation to propagate.
+                        catch (Exception ex) when (ex as OperationCanceledException == null)
```
pranavkm: Is this easier to grock: `when (!(ex is OperationCanceledException))`

test/Microsoft.Extensions.Diagnostics.HealthChecks.Tests/DefaultHealthCheckServiceTest.cs
```diff
@@ -188,6 +188,35 @@ public async Task CheckHealthAsync_SetsRegistrationForEachCheck()
...
+            // Arrange
+            var insideCheck = new TaskCompletionSource<object>();
+
+            var service = CreateHealthChecksService(b =>
+            {
+                b.AddAsyncCheck("cancels", async ct =>
+                {
+                    insideCheck.SetResult(null);
+
+                    await Task.Delay(10000, ct);
```
pranavkm: Could you use a `ManualResetEvent` here?

rynowak: I could do a lot of things, why would I want to do that?

src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
rynowak: eh, I like it the way it is

test/Microsoft.Extensions.Diagnostics.HealthChecks.Tests/DefaultHealthCheckServiceTest.cs
rynowak: Discussed offline, the delay doesn't actually happen because cancellation happens right away

PR #485, opened 2018-09-23
Set cache headers in health check middleware

rynowak:


src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -70,6 +71,13 @@ public async Task InvokeAsync(HttpContext httpContext)
 
             httpContext.Response.StatusCode = statusCode;
 
+            if (!_healthCheckOptions.SuppressCacheHeaders)
```
rynowak: I considered also a design where you could give us the value you wanted to return from `Cache-Control`, but needing to support `Pragma` complicates things.

Some h/c systems are HTTP 1.0 so this is an area where we **do** care.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -70,6 +71,13 @@ public async Task InvokeAsync(HttpContext httpContext)
...
+            {
+                var headers = httpContext.Response.Headers;
+                headers[HeaderNames.CacheControl] = "no-store, no-cache";
+                headers[HeaderNames.Pragma] = "no-cache";
```
Tratcher: [Elsewhere](https://github.com/aspnet/Security/blob/7b6c9cf0eeb149f2142dedd55a17430e7831ea99/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L377-L379) we've also done Expires for thoroughness.

rynowak:
Updated

rynowak:
@Tratcher - I noticed the linked code uses `no-cache` and we're using `no-cache, no-store` (based on some places in MVC) - which should we be doing here?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -16,7 +17,7 @@ public class HealthCheckMiddleware
         private readonly RequestDelegate _next;
         private readonly HealthCheckOptions _healthCheckOptions;
         private readonly HealthCheckService _healthCheckService;
-
+
```
davidfowl: nit: whitespace

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/Microsoft.AspNetCore.Diagnostics.HealthChecks.csproj
```diff
@@ -18,6 +18,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="$(MicrosoftAspNetCoreHttpAbstractionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="$(MicrosoftExtensionsOptionsPackageVersion)" />
+    <PackageReference Include="Microsoft.Net.Http.Headers" Version="$(MicrosoftNetHttpHeadersPackageVersion)" />
```
davidfowl: Why do you need this?

rynowak: `HeaderNames` is defined there for some reason - is there a problem with using it?

davidfowl: Wow I had no idea we put it there. In general, I'd like this package to die a slow death. We need to re-unify with the real typed headers in .NET Core. Anyways, it's fine to have this for now I guess.

rynowak: Yeah 😞 I spent a while looking for HeaderNames because I thought it surely couldn't be defined in *that* package.

test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
```diff
@@ -291,6 +292,57 @@ public async Task CanSetCustomStatusCodes()
...
+            var server = new TestServer(builder);
+            var client = server.CreateClient();
+
+            var response = await client.GetAsync("/health");
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("Healthy", await response.Content.ReadAsStringAsync());
+            Assert.Equal("no-store, no-cache", response.Headers.CacheControl.ToString());
+            Assert.Equal("no-cache", response.Headers.Pragma.ToString());
+            Assert.Equal("Thu, 01 Jan 1970 00:00:00 GMT", response.Content.Headers.Expires.ToString());
```
rynowak: So this ToStrings by using culture sensitive formatting instead of giving the original text.... It's fun when writing the unit tests is 5x harder than writing the actual codez

rynowak: #TypedHeaders

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
```diff
@@ -1,13 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Http;
```
pranavkm: Tsck tsck

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckOptions.cs
```diff
@@ -46,5 +46,13 @@ public class HealthCheckOptions
         /// of <see cref="HealthReport.Status"/> as a string.
         /// </remarks>
         public Func<HttpContext, HealthReport, Task> ResponseWriter { get; set; } = HealthCheckResponseWriters.WriteMinimalPlaintext;
+
+        /// <summary>
+        /// Gets or sets a value that controls whether the health check middleware will add HTTP headers to prevent
+        /// response caching. If the value is <c>false</c> the health check middleware will set the <c>Cache-Control</c>,
```
pranavkm: > will set or override existing values for ...

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
pranavkm: using System* before everything else

Tratcher:
@rynowak here's a good [description](https://stackoverflow.com/questions/866822/why-both-no-cache-and-no-store-should-be-used-in-http-response). Your version should be fine.

PR #487, opened 2018-09-23
Add EFCore DbContext check

rynowak:
First draft of a DbContext-based EFCore health check - the sample is probably the best place to start.


rynowak:
> but looking at this I wonder if you really need to use IOptionsMonitor.

I thought IOptionsMonitor was the way to do named options? 🤷‍♂️

divega:
> > but looking at this I wonder if you really need to use IOptionsMonitor.
> 
> I thought IOptionsMonitor was the way to do named options? 🤷‍♂️

Maybe. As I said, I don’t remember enough. @haok?

HaoK:
Yeah no other way to do named options other than `IOptionsMonitor`, looks fine to me

rynowak:
Cool, thanks friends!

rynowak:
Will need to wait to add packages to aspnet/Universe before merging 😲

Issue #491, opened 2018-09-27
[Feature request] HealthCheckResult and health checks execution time


unaizorrilla:
There is any plan to include the execution time ( or the whole execution time ) of health checks in the HealthCheckResult?

This data can be usefull for other stories like HealthCheck tracking results.
PR can be accepted?

mkArtakMSFT:
Thanks for contacting us, @unaizorrilla.
@glennc, @rynowak thoughts? Including the `execution time` into the result seems like a reasonable ask to me. I would actually even think that the execution time would be useful per `HealthReportEntry`.

rynowak:
Sure, we can this, this seems reasonable.

unaizorrilla:
PR can be accepted?

rynowak:
Sure

rynowak:
Feel free to send a PR - if you don't get it together by lets say Weds next week then I will just do it so that it gets into preview3.

What I had in mind for this was to add `public TimeSpan Duration { get; }` property to `HealthReportEntry` and a `public TimeSpan TotalDuration { get; }` to `HealthReport`.

We already collect this data for logging, so this will be pretty easy:
https://github.com/aspnet/Diagnostics/blob/release/2.2/src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs#L49
https://github.com/aspnet/Diagnostics/blob/release/2.2/src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs#L83

rynowak:
Thanks!

PR #493, opened 2018-09-28
Added execution time duration (HealthReportEntry TotalDuration)

unaizorrilla:
Hi @rynowak 

This is the PR related with the issue #491. 

HealthReport total duration is setted using the totaltime you mentioned on the issue, my only doubt is whether this time should not be aggregated from the report entry times.

Thanks

### DIFF
--- a/src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
+++ b/src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
@@ -73,15 +73,16 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks
                         try
                         {
                             var result = await healthCheck.CheckHealthAsync(context, cancellationToken);
+                            var duration = stopwatch.GetElapsedTime();
 
                             entry = new HealthReportEntry(
                                 status: result.Result ? HealthStatus.Healthy : registration.FailureStatus,
                                 description: result.Description,
-                                duration: stopwatch.GetElapsedTime(),
+                                duration: duration,
                                 exception: result.Exception,
                                 data: result.Data);
 
-                            Log.HealthCheckEnd(_logger, registration, entry, stopwatch.GetElapsedTime());
+                            Log.HealthCheckEnd(_logger, registration, entry, duration);
                             Log.HealthCheckData(_logger, registration, entry);
                         }
 
--- a/src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
+++ b/src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
@@ -75,10 +75,11 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks
                             var result = await healthCheck.CheckHealthAsync(context, cancellationToken);
 
                             entry = new HealthReportEntry(
-                                result.Result ? HealthStatus.Healthy : registration.FailureStatus,
-                                result.Description,
-                                result.Exception,
-                                result.Data);
+                                status: result.Result ? HealthStatus.Healthy : registration.FailureStatus,
+                                description: result.Description,
+                                duration: stopwatch.GetElapsedTime(),
+                                exception: result.Exception,
+                                data: result.Data);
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthReportEntry.cs
```diff
@@ -19,12 +19,14 @@ public struct HealthReportEntry
         /// </summary>
         /// <param name="status">A value indicating the health status of the component that was checked.</param>
         /// <param name="description">A human-readable description of the status of the component that was checked.</param>
+        /// <param name="duration">A value indicating the health execution duration.</param>
         /// <param name="exception">An <see cref="Exception"/> representing the exception that was thrown when checking for status (if any).</param>
         /// <param name="data">Additional key-value pairs describing the health of the component.</param>
-        public HealthReportEntry(HealthStatus status, string description, Exception exception, IReadOnlyDictionary<string, object> data)
+        public HealthReportEntry(HealthStatus status, string description,TimeSpan duration, Exception exception, IReadOnlyDictionary<string, object> data)
```
ilmax: nit: missing space

src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
```diff
@@ -75,10 +75,11 @@ internal class DefaultHealthCheckService : HealthCheckService
                             var result = await healthCheck.CheckHealthAsync(context, cancellationToken);
 
                             entry = new HealthReportEntry(
-                                result.Result ? HealthStatus.Healthy : registration.FailureStatus,
-                                result.Description,
-                                result.Exception,
-                                result.Data);
+                                status: result.Result ? HealthStatus.Healthy : registration.FailureStatus,
+                                description: result.Description,
+                                duration: stopwatch.GetElapsedTime(),
```
rynowak: I think it would be good to store `stopwatch.GetElapsedTime()` in a local here because it's used twice (L80 & L84). Normally I wouldn't care about this, but I want to avoid logging + health report entry having different time values.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthReport.cs
```diff
@@ -14,10 +16,12 @@ public sealed class HealthReport
         /// Create a new <see cref="HealthReport"/> from the specified results.
         /// </summary>
         /// <param name="entries">A <see cref="IReadOnlyDictionary{TKey, T}"/> containing the results from each health check.</param>
-        public HealthReport(IReadOnlyDictionary<string, HealthReportEntry> entries)
+        /// <param name="totalDuration">A value indicating the total duration of executing <paramref name="entries"/>.</param>
```
rynowak: Suggestion (use the wording your used on the property):

> A value indicating the total duration of executing a group of <see cref="IHealthCheck"/> instances.

I think it's good that we define this as the time the health check service took to execute.

rynowak:
> HealthReport total duration is setted using the totaltime you mentioned on the issue, my only doubt is whether this time should not be aggregated from the report entry times.

I think defining the total time as "the time the health check service took to execute" vs "the sum of all health check timings" is the right decision. This will help users keep us honest, and if we manage to introduce significant overheads in this service it will be obvious.

src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
```diff
@@ -73,30 +73,39 @@ internal class DefaultHealthCheckService : HealthCheckService
...
                         catch (Exception ex) when (ex as OperationCanceledException == null)
                         {
-                            entry = new HealthReportEntry(HealthStatus.Failed, ex.Message, ex, data: null);
+                            entry = new HealthReportEntry(
+                                status: HealthStatus.Failed,
+                                description: ex.Message,
+                                duration: stopwatch.GetElapsedTime(),
+                                exception: ex,
+                                data: null);
+
```
rynowak: Sorry, forgot to comment here as well, this should also use a local for the time (L95 & L99)

PR #496, opened 2018-10-06
Adjust log levels

rynowak:


src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
```diff
@@ -144,10 +144,28 @@ public static class EventIds
...
-                "Health check {HealthCheckName} completed after {ElapsedMilliseconds}ms with status {HealthStatus} and '{HealthCheckDescription}'");
+                HealthCheckEndText);
+
+            private static readonly Action<ILogger, string, double, HealthStatus, string, Exception> _healthCheckEndDegraded = LoggerMessage.Define<string, double, HealthStatus, string>(
+                LogLevel.Warning,
+                EventIds.HealthCheckEnd,
+                HealthCheckEndText);
+
+            private static readonly Action<ILogger, string, double, HealthStatus, string, Exception> _healthCheckEndUnhealthy = LoggerMessage.Define<string, double, HealthStatus, string>(
+                LogLevel.Error,
```
pranavkm: Tangentially related, but could we add a remark to `HealthStatus.Unhelathy` and say that it's typically the value returned by default if a health check does not report a status? It's not entirely clear what the difference is between `Unhealthy` and `Degraded` and I wasn't sure why this warranted an error rather than a warning.

src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
```diff
@@ -171,7 +189,25 @@ public static void HealthCheckBegin(ILogger logger, HealthCheckRegistration regi
...
+
+                    case HealthStatus.Unhealthy:
+                        _healthCheckEndUnhealthy(logger, registration.Name, duration.TotalMilliseconds, entry.Status, entry.Description, null);
+                        break;
+
+                    case HealthStatus.Failed:
+                        _healthCheckEndFailed(logger, registration.Name, duration.TotalMilliseconds, entry.Status, entry.Description, null);
+                        break;
+                }
+
```
pranavkm: Extra line

src/Microsoft.Extensions.Diagnostics.HealthChecks/DefaultHealthCheckService.cs
rynowak: > returned by default if a health check does not report a status

I'm not sure what that means. Health checks are pass/fail. We allow you to customize the severity of `fail`

rynowak: I think I know what you mean after all. Adding some text to the docs.

PR #498, opened 2018-10-07
Add IHealthCheckPublisher for push-based checks

rynowak:
IHealthCheckPublisher allows you to configure and run health checks
regularly inside an application, and push the notifications elsewhere.

All publishers are part of a single queue with a configurable period and
timeout.


src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/IHealthCheckPublisher.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// Publishes the provided <paramref name="report"/>.
+        /// </summary>
+        /// <param name="report">The <see cref="HealthReport"/>. The result of executing a set of health checks.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/>.</param>
+        /// <returns>A <see cref="Task"/> which will complete when publishing is complete.</returns>
+        Task PublishAsync(HealthReport report, CancellationToken cancellationToken);
+    }
+}
```
rynowak: /cc @unaizorrilla

unaizorrilla: ack

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
```diff
@@ -0,0 +1,269 @@
...
+        {
+            var duration = ValueStopwatch.StartNew();
+            Logger.HealthCheckPublisherProcessingBegin(_logger);
+
+            CancellationTokenSource cancellation = null;
+            try
+            {
+                var timeout = _options.Value.Timeout;
+
+                cancellation = CancellationTokenSource.CreateLinkedTokenSource(_stopping.Token);
```
rynowak: This part isn't thread safe enough, I need to be careful accessing `_stopping`.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Gets or sets a predicate that is used to filter the set of health checks executed.
+        /// </summary>
+        /// <remarks>
+        /// If <see cref="Predicate"/> is <c>null</c>, the health check publisher service will run all
+        /// registered health checks - this is the default behavior. To run a subset of health checks,
+        /// provide a function that filters the set of checks.
+        /// </remarks>
+        public Func<HealthCheckRegistration, bool> Predicate { get; set; }
```
Tratcher: Filter?

Tratcher: Does this need to run every Period vs running once at startup? The comment should clarify the run frequency regardless.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
Tratcher: Don't null out or Dispose _stopping, Cancel is more than adequate and thread safe.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
```diff
@@ -0,0 +1,269 @@
...
+                if (timeout != Timeout.InfiniteTimeSpan)
+                {
+                    cancellation.CancelAfter(timeout);
```
Tratcher: CancelAfter understands InfiniteTimeSpan, you don't need an explicit check.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
```diff
@@ -0,0 +1,269 @@
...
+                throw new ArgumentNullException(nameof(publishers));
+            }
+
+            _healthCheckService = healthCheckService;
+            _options = options;
+            _logger = logger;
+            _publishers = publishers.ToArray();
+        }
+
+        internal bool IsCancelled => _stopping == null || _stopping.IsCancellationRequested;
```
Tratcher: Yeah, more bad races, don't null out _stopping

Tratcher: IsStopping

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
```diff
@@ -0,0 +1,269 @@
...
+            var tasks = new Task[publishers.Length];
+            for (var i = 0; i < publishers.Length; i++)
+            {
+                try
+                {
+                    tasks[i] = RunPublisherAsync(publishers[i], report, cancellationToken);
+                }
+                catch (Exception ex)
+                {
+                    // Handle any exceptions that are thrown synchronously - this allows us to start
```
Tratcher: async/await methods don't throw exceptions synchronously, they capture all exceptions and throw them later.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// Options for the default service that executes <see cref="IHealthCheckPublisher"/> instances.
+    /// </summary>
+    public sealed class HealthCheckPublisherOptions
+    {
+        /// <summary>
+        /// Gets or sets the initial delay applied after the application starts before executing 
+        /// <see cref="IHealthCheckPublisher"/> instances. The delay is applied once at startup, and does
+        /// not apply to subsequent iterations. The default value is 5 seconds.
+        /// </summary>
+        public TimeSpan Delay { get; set; } = TimeSpan.FromSeconds(5);
```
Tratcher: Valid range? What if I set 0 or InfinateTimeSpan?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Gets or sets the period of <see cref="IHealthCheckPublisher"/> execution. The default value is
+        /// 30 seconds.
+        /// </summary>
+        public TimeSpan Period { get; set; } = TimeSpan.FromSeconds(30);
```
Tratcher: Valid range? What if I set 0 or InfinateTimeSpan?

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
rynowak: So what does dispose do then? I thought we were worried about subscriptions to the cancellation token capturing state?

Tratcher: Dispose is only necessary for tokens that will not be cancelled. Cancel cleans up all state and new registrations are invoked inline.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
rynowak: Cool beans 👍

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
rynowak: Yeah, I think this here from before I wrote `RunPublisherAsync`. This code used to call the publisher directly.

I can't imagine a world where we make `RunPublisherAsync` into a sync implementation - so I'm ok with this change.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
rynowak: Cool, good to know!

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherOptions.cs
rynowak: Keeping the name predicate as we are using that elsewhere

src/Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions/IHealthCheckPublisher.cs
rynowak: OK, this is going to go in sometime soon, please let us know if you have any feedback because it will be hard to change in about a week 😆

rynowak:
@Tratcher Updated

samples/WelcomePageSample/web.config
```diff
@@ -4,6 +4,8 @@
     <handlers>
```
Tratcher: Why is this file even here? It's not needed.

src/Microsoft.Extensions.Diagnostics.HealthChecks/HealthCheckPublisherHostedService.cs
```diff
@@ -77,6 +77,8 @@ public Task StartAsync(CancellationToken cancellationToken = default)
 
         public Task StopAsync(CancellationToken cancellationToken = default)
         {
+            _stopping.Cancel();
```
Tratcher: try/catch Exception, callbacks are invoked synchronously and can throw.

samples/WelcomePageSample/web.config
rynowak: OK

PR #501, opened 2018-10-10
Adjust font-weight to make it easier to read

06b:
Personally I've found the developer exception page can be hard to read with the really thin font-weight even with 20/20 vision.

Below is an example of the default on a 1.5 device pixel ratio screen for chrome, edge & firefox.
chromeedgefirefox-1_5dppx-default

Below is a screenshot with what I'm suggesting with this pull request. 
chromeedgefirefox-1_5dppx-code_fontweight600-h1h2h3h4h5_fontweight400

Obviously with different people, devices, browsers, fonts, settings, etc - this can be considered subjective, so if you think I should create an issue on the Home repo to get other people's feedback, let me know.

I also removed the commented out declaration.

### DIFF
--- a/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
+++ b/src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
@@ -6,8 +6,7 @@
 }
 
 h1, h2, h3, h4, h5 {
-    /*font-family: 'Segoe UI',Tahoma,Arial,Helvetica,sans-serif;*/
-    font-weight: 100;
+    font-weight: 400;
 }
 
 h1 {
@@ -26,6 +25,7 @@ h3 {
 
 code {
     font-family: Consolas, "Courier New", courier, monospace;
+    font-weight: 600;
 }
 
 body .titleerror {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
```diff
@@ -6,8 +6,7 @@
 }
 
 h1, h2, h3, h4, h5 {
-    /*font-family: 'Segoe UI',Tahoma,Arial,Helvetica,sans-serif;*/
-    font-weight: 100;
+    font-weight: 400;
```
Eilon: Out of curiosity, what if you just leave this out entirely (and the one below)? Sometimes the problem is that *too many* things are specified.

Eilon:
This does look a lot nicer! I could never pinpoint why I found it so hard to read the stack traces... but it's those thin, wiry fonts!

src/Microsoft.AspNetCore.Diagnostics/DeveloperExceptionPage/Views/ErrorPage.css
Eilon: This screen shot looks pretty nice to me. Maybe we give that a go?

Eilon (approved):

Eilon:
@mkArtakMSFT can you assign someone to review & merge? I'm a fan of this change.

mccoyrjm:
Much easier to read, supporting this!

natemcmaster:
This looks like something we should get merged. The code for this project has moved to https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/Diagnostics/src/DeveloperExceptionPage/Views/ErrorPage.css as a part of merging repos. Can you re-open this PR against the new location?

Thanks,
Nate

Issue #511, opened 2018-10-23
Nested paths screw up health check filters


jmezach:
I'm not sure if this is even supposed to work, but it's just something I noticed while I was exploring the health checks feature. In my ConfigureServices I have the following:

```csharp
services.AddHealthChecks()
        .AddCheck("MyDatabase",
                  new SqlConnectionHealthCheck("Data Source=localhost;Initial Catalog=my-dummy-database"),
                  tags: new[] { "ready" });
```

Obviously the SqlConnectionHealthCheck fails because it cannot connect to the database (in fact, there isn't even a SQL Server running on my machine). Now I have added the following to my Configure method:

```csharp
app.UseHealthChecks("/health", new HealthCheckOptions {});

app.UseHealthChecks("/health/ready", new HealthCheckOptions
{
  Predicate = (check) => check.Tags.Contains("ready"), 
});

app.UseHealthChecks("/health/live", new HealthCheckOptions()
{
  Predicate = (check) => false,
});
```

In this setup I would expect **/health** to return **Unhealthy**, since it runs the SqlConnectionHealthCheck and it fails. This is indeed the case. The same goes for the **/health/ready** endpoint. However, I was expecting **/health/live** to return **Healthy**, but it does not. Instead it returns **Unhealthy** and looking at the log it seems to be executing the SqlConnectionHealthCheck regardless of the predicate.

Personally I would be fine with it if this is something that's not supported, but it might be useful to put this into the documentation somewhere so that users of this feature don't get tripped up.

rynowak:
Thanks for the feedback. This is using the `.Map` middleware prefix which matches any suffix of the path you put in. 

However seeing your code, I agree that this is really surprising. We'll look at changing this.

mkArtakMSFT:
@rynowak, can you please assign a priority to this? Thanks!

rynowak:
Fixed for the 2.2.0 release. Thanks for the feedback 👍

jmezach:
Nice, looking forward to trying out the new bits.

Issue #514, opened 2018-10-26
Cannot configure multiple paths over same custom port


herrirya:
It appears it is not possible to have multiple paths that use the same custom port number (which is likely unintentional). For example:

```
app.UseHealthChecks("/health/live", port: 9999);
app.UseHealthChecks("/health/ready", port: 9999);
```

The second configured path in the example ("/health/ready") is not accessible because the first configured path is hogging up the port - seemingly due to the equality condition set up in HealthCheckApplicationBuilderExtensions.cs UseHealthChecksCore() method, line 214.

```
app.MapWhen(
  c => c.Connection.LocalPort == port,
  b0 => b0.Map(path, b1 => b1.UseMiddleware<HealthCheckMiddleware>(args)));
```

This middleware is great; thank you for your work.

rynowak:
Thanks. This looks like this is indeed a bug.

rynowak:
Fixed for the 2.2.0 release. Thanks for the feedback 👍

PR #516, opened 2018-10-28
Don't use Map

rynowak:
Fixes aspnet/Diagnostics#511 and aspnet/Diagnostics#514

It's really confusing to people that we use Map. Users expect that the
URL they provide for the health check middleware will only process
exact matches. The way it behaves when using map is not optimal for some
of the common patterns.

### DIFF
--- a/test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
+++ b/test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
@@ -396,6 +396,131 @@ namespace Microsoft.AspNetCore.Diagnostics.HealthChecks
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
```diff
@@ -395,6 +395,131 @@ public async Task CanListenWithoutPath_AcceptsRequest()
             Assert.Equal("Healthy", await response.Content.ReadAsStringAsync());
         }
 
+        [Fact]
+        public async Task CanListenWithPath_AcceptsRequestWithExtraSlash()
```
rynowak: Adding some additional tests here because we're not using `.Map` anymore, we're doing a semi-custom thing.

rynowak:
👍

Issue #509, opened 2018-10-21
Consider changing the option name related to cache headers


rynowak:
See discussion [here](https://github.com/aspnet/Docs/pull/8959#discussion_r226867887) - I think we could choose a better name for this

PR #517, opened 2018-10-28
Rename anti-caching option

rynowak:
- Renamed the property for configuration response caching headers
- Renamed the options class to avoid conflicts with other type names

Fixes https://github.com/aspnet/Diagnostics/issues/509

### DIFF
--- a/test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
+++ b/test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
@@ -324,7 +324,7 @@ namespace Microsoft.AspNetCore.Diagnostics.HealthChecks
                 {
                     app.UseHealthChecks("/health", new HealthCheckOptions()
                     {
-                        SuppressCacheHeaders = true,
+                        AllowCachingResponses = true,
                     });
                 })
                 .ConfigureServices(services =>
--- a/src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddleware.cs
@@ -71,7 +71,7 @@ namespace Microsoft.AspNetCore.Diagnostics.HealthChecks
 
             httpContext.Response.StatusCode = statusCode;
 
-            if (!_healthCheckOptions.SuppressCacheHeaders)
+            if (!_healthCheckOptions.AllowCachingResponses)
             {
                 // Similar to: https://github.com/aspnet/Security/blob/7b6c9cf0eeb149f2142dedd55a17430e7831ea99/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L377-L379
                 var headers = httpContext.Response.Headers;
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddlewareOptions.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.AspNetCore.Diagnostics.HealthChecks
     /// <summary>
     /// Contains options for the <see cref="HealthCheckMiddleware"/>.
     /// </summary>
-    public class HealthCheckOptions
+    public class HealthCheckMiddlewareOptions
```
pranavkm: This is different from the scheme we use in other middleware e.g. https://github.com/aspnet/StaticFiles/blob/master/src/Microsoft.AspNetCore.StaticFiles/StaticFileOptions.cs. Intentional?

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddlewareOptions.cs
```diff
@@ -48,11 +48,19 @@ public class HealthCheckOptions
...
 
         /// <summary>
-        /// Gets or sets a value that controls whether the health check middleware will add HTTP headers to prevent
-        /// response caching. If the value is <c>false</c> the health check middleware will set or override the 
+        /// Gets or sets a value that controls whether responses from the health check middleware can be cached.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The health check middleware does not perform caching of any kind. This setting configures whether
+        /// the middleware will apply headers to the HTTP response that instruct clients to prevent caching.
```
pranavkm: that ~~instruct~~ instructs clients

test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
```diff
@@ -322,9 +322,9 @@ public async Task CanSuppressCacheHeaders()
             var builder = new WebHostBuilder()
                 .Configure(app =>
                 {
-                    app.UseHealthChecks("/health", new HealthCheckOptions()
+                    app.UseHealthChecks("/health", new HealthCheckMiddlewareOptions()
                     {
-                        SuppressCacheHeaders = true,
+                        AllowCachingResponses = true,
```
pranavkm: `AllowCachingResponses  = false`?

pranavkm (approved):

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddlewareOptions.cs
pranavkm: Nevermind, just saw your comment. What other type is this conflicting with?

rynowak: Yeah this was kinda speculative on my part... will undo.

src/Microsoft.AspNetCore.Diagnostics.HealthChecks/HealthCheckMiddlewareOptions.cs
rynowak: no, this is already correct.

test/Microsoft.AspNetCore.Diagnostics.HealthChecks.Tests/HealthCheckMiddlewareTests.cs
rynowak: no, this is correct. It prevents the cache headers from being added. The property was renamed, but it's purpose and default value are the same.

PR #527, opened 2018-11-13
Switch async void to async Task

javiercn:


ryanbrandenburg (approved):
It's my understanding that this is always preferable.

pranavkm (approved):

